PR #2, opened 2021-08-26
OEL-306: Provide default content feature.

escuriola:
### Description

Added required features to allow to store default content for showcase library.

### Change log

- Added: oe_showcase_default_content submodule
- Changed:
- Deprecated:
- Removed: 
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/modules/oe_showcase_default_content/README.md
+++ b/modules/oe_showcase_default_content/README.md
@@ -1,0 +1,3 @@
+# OpenEuropa Showcase default content
+
+Provide default content to oe_showcase pages
--- a/composer.json
+++ b/composer.json
@@ -18,6 +18,7 @@
         "openeuropa/code-review": "^1.6",
         "openeuropa/drupal-core-require-dev": "^8.9 || ^9.1",
         "openeuropa/task-runner-drupal-project-symlink": "^1.0",
+        "drupal/default_content": "^2.0@alpha",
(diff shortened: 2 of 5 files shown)


modules/oe_showcase_default_content/README.md
```diff
@@ -0,0 +1,3 @@
+# OpenEuropa Showcase default content
+
+Provide default content to oe_showcase pages
```
GilNovacomm, line 3: The rationale here is that in the context of oe_showcase this module allows other modules to provide default content not oe_showcase_default_content directly.
So maybe update the description to reflect this idea.

GilNovacomm (commented):
Consider rewriting the README file description.
This module doesn't directly provides default content, in the context of oe_showcase it allows other modules to provide default content at installation.

modules/oe_showcase_default_content/README.md
escuriola: Not exactly. All the content will be stored in this module, as explained in the ticket.

GilNovacomm: Ok I've confirmed with Andras that this module will also provide default content if this content is reused by other features so the description also applies.

GilNovacomm (dismissed):

drishu (approved):

GilNovacomm (approved):

PR #4, opened 2021-08-30
OEL-302: Create default content for oe_showcase_page

escuriola:
### Description

Added required features to allow to store default content for showcase library.

### Change log

- Added: default content for oe_showcase_page module
- Changed:
- Deprecated:
- Removed: 
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


modules/oe_showcase_page/oe_showcase_page.info.yml
```diff
@@ -0,0 +1,28 @@
...
+config_devel:
+  install:
+    - core.entity_form_display.node.oe_showcase_page.default
+    - core.entity_view_display.node.oe_showcase_page.teaser
+    - core.entity_view_display.node.oe_showcase_page.default
+    - field.field.node.oe_showcase_page.field_description
+    - field.field.node.oe_showcase_page.field_body
+    - field.storage.node.field_description
+    - field.storage.node.field_body
+    # - editor.editor.full_html
```
abel-santos-corral: If not used, better eliminate it.

abel-santos-corral (changes requested):
Added comments at the PR regarding the menus label must be equal to title. Some authoring info could be hidden, and Block title could be added. To be checked.

drishu:
> Added comments at the PR regarding the menus label must be equal to title.
Yes
> Some authoring info could be hidden,
No, not yet, we don't have specs for that, we will see in the review demo
> and Block title could be added. To be checked.
which block ?

abel-santos-corral:
@drishu the block would be Page title block. So when visiting an OE page, to see the title. For now, even for search content type, when visiting a node, the title is not shown.

drishu:
> @drishu the block would be Page title block. So when visiting an OE page, to see the title. For now, even for search content type, when visiting a node, the title is not shown.

Agreed

drishu (commented):
- Place the title for the Page content type, but not for <front>
- Lets use OpenEuropa instead of Openeuropa or Open Europa everywhere
- Remove the body from the homepage
- For the content type lets hide the labels of the fields
- Fix the breadcrumbs, it should not say "Node" in the middle, it should take its hierarchy from the menu items (breadcrumbs config)
- Use below for the homepage body:
<p>&nbsp;</p>
<p style="text-align:center;">This profile is aimed to provide developers with a series of examples on how the Open Europa Library features are built and configured.&nbsp;Exposing these features,&nbsp;the building of new sites will be easy to get the necessary components to achieve every site's goals.</p>
<p>&nbsp;</p>
<p>&nbsp;</p>
- on Rich text and documents, lets add "Documents:" in bold right before the documents (at the bottom of the richtext)
- lets move the first 2 menu items: "Rich Text" and "Rich Text with Documents" outside pages

drishu (dismissed):

drishu (approved):

GilNovacomm (approved):
Approved.

abel-santos-corral (approved):

PR #6, opened 2021-09-02
OEL-257: Add Search API Autocomplete to Search View

GilNovacomm:


runner.yml.dist
```diff
@@ -22,6 +22,7 @@ drupal:
     - "./vendor/bin/drush -y config-set system.performance css.preprocess 0"
     - "./vendor/bin/drush -y config-set system.performance js.preprocess 0"
     # Enable the modules.
+    - "./vendor/bin/drush en oe_showcase_default_content -y"
```
cmarcmar: Shouldn't we enable modules in the config/sync/core.extension.yml file instead?

drishu: in profiles we should not export that

Maxfire (dismissed):
reviewed and works fine.

drishu (approved):

Maxfire (approved):
approved :+1:

PR #10, opened 2021-09-15
OEL-447: Integrate oe_whitelabel into oe_showcase

abel-santos-corral:
### Description

Once oe_whitelabel has an alpha release we will need to add it as default theme of the oe_showcase profile.

- Add oe_whitelabel to composer.json if there are any images in whitelabel that are not present in showcase add them to the .drone.yml and docker-compose.yml files
- Change runner.yml to enable oe_whitelabel and submodules
- Place blocks provided by the whitelabel and its dependencies
- See analysis in confluence, the breadcumbs need easy_breadcrumb added and configured, the block placed beneath blue bar (that is `header` region)
- We need behat test to asset the presence of the block and navigation elements (see analysis scenarios)

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


abel-santos-corral:
At commit https://github.com/openeuropa/oe_showcase/pull/10/commits/0b9620af4494d0fd5aa62d2dd45d741d50cb55a2 introduced changes of 
- info.ym lfile
- runner
After comparing with other profiles (standard, minimal, etc) we realized that in info.yml file instead of dependencies we should state install. We added toolbar.
From runner we removed the toolbar module and we eliminate the activation of several modules per line to set an activation per line.

drishu (approved):

escuriola (approved):

PR #16, opened 2021-09-23
OEL-000: Better badge styling in card.

drishu:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/modules/oe_showcase_search/templates/node--search-result.html.twig
+++ b/modules/oe_showcase_search/templates/node--search-result.html.twig
@@ -8,8 +8,11 @@
 {% if content.field_category is not empty %}
   {% for key,item in content.field_category %}
     {% if key|first != '#' %}
+      {% set badge_attr = create_attribute() %}
+      {% set badge_attr = badge_attr.addClass('me-2') %}
       {% set badges = badges|merge([{
-        label: item['#title']
+        label: item['#title'],
+        attributes: badge_attr
       }]) %}
     {% endif %}
   {% endfor %}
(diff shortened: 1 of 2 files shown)


modules/oe_showcase_search/templates/node--search-result.html.twig
```diff
@@ -8,8 +8,11 @@
 {% if content.field_category is not empty %}
   {% for key,item in content.field_category %}
     {% if key|first != '#' %}
+      {% set badge_attr = create_attribute() %}
+      {% set badge_attr = badge_attr.addClass('me-2') %}
```
Maxfire: - I guess same defined https://github.com/openeuropa/oe_whitelabel/pull/22/files#r716473197 could be applied here.
- btw why setting the badge in the for loop?, I think we could create the attribute before the for loop and add it in the badge variable afterwards, so that we're not creating the badge for every iteration in the loop.

Maxfire (changes requested):

runner.yml.dist
```diff
@@ -29,6 +29,8 @@ drupal:
     - "./vendor/bin/drush config-set system.theme default oe_whitelabel -y"
     - "./vendor/bin/drush config-set system.theme admin seven -y"
     - "./vendor/bin/drush config-set node.settings use_admin_theme 1 -y"
+    # Import translations.
```
abel-santos-corral: Better put activation of oe_multilingual and afterwards import translations, and eliminate this comment line that doesn't add value.

abel-santos-corral (approved):

Maxfire (approved):

PR #27, opened 2021-10-15
OEL-579: Showcase - Medley branch BCL 0.9.0 with all components upgraded.

escuriola:
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-579

### DIFF
--- a/modules/oe_showcase_search/templates/node--search-result.html.twig
+++ b/modules/oe_showcase_search/templates/node--search-result.html.twig
@@ -5,7 +5,8 @@
  */
 #}
 
-{{ pattern('card_search', {
+{{ pattern('card', {
+  variant: 'search',
   title: label,
   text: content.body,
   image: content.field_image[0],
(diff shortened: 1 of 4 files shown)


modules/oe_showcase_search/templates/node--search-result.html.twig
```diff
@@ -10,20 +10,22 @@
   {% for key,item in content.field_category %}
```
drishu: let's see if we can prep a variant of the card pattern inside whitelabel that offers these configurations and parsing out of the box

drishu (changes requested):

modules/oe_showcase_search/templates/node--search-result.html.twig
escuriola: Not done yet.

escuriola: Now it's done.

abel-santos-corral (approved):

abel-santos-corral (approved):

timpevi (approved):

drishu (approved):

PR #30, opened 2021-10-20
OEL-520: Add page integration test

GilNovacomm:
[OEL-520](https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-520)

### Changes

Config changes:
- Allow `oe_showcase_page.field_body` to reference `oe_document` paragraphs.
- No longer allow `oe_showcase_page.field_body` to reference `oe_timeline` paragraphs.
  This reverts part of OEL-998, see modules/oe_showcase_page/config/install/field.field.node.oe_showcase_page.field_body.yml

New functional test:
- Create new oe_showcase_page node through the UI (node/add/*).
- Set all or most of the fields, and add various example paragraphs in field_body.
- Visit node page, and verify that all values are displayed.

### Merge notes
When merging, ideally we should have separate commits for the config changes and the new test.
As this is not easily possible through the UI, please include the "Changes" part from the PR text in the merge commit message.

### DIFF
--- a/modules/oe_showcase_page/tests/src/ExistingSite/PageTest.php
+++ b/modules/oe_showcase_page/tests/src/ExistingSite/PageTest.php
@@ -1,0 +1,141 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\Tests\oe_showcase_page\ExistingSite;
+
+use Drupal\Tests\oe_showcase\ExistingSite\ShowcaseExistingSiteTestBase;
+
+/**
+ * Tests the 'oe_showcase_page' content type.
+ */
+class PageTest extends ShowcaseExistingSiteTestBase {
+
+  /**
+   * A user with permission to create 'oe_showcase_page' content.
+   *
+   * @var \Drupal\user\UserInterface
+   */
+  protected $user;
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function setUp(): void {
+    parent::setUp();
+
+    $permissions = [
+      'create oe_showcase_page content',
+    ];
+
+    $this->user = $this->createUser($permissions);
+  }
+
+  /**
+   * Create test Showcase Page content.
+   */
+  public function testCreateShowCasePage() {
+    // Mark test content for deletion after the test has finished.
+    $this->markEntityTypeForCleanup('node');
+    $this->markEntityTypeForCleanup('paragraph');
+
+    $assert_session = $this->assertSession();
+    $this->drupalLogin($this->user);
+
+    // Create a Showcase Page node through the UI.
+    $this->drupalGet('node/add/oe_showcase_page');
+    $page = $this->getSession()->getPage();
+
+    // Set page title.
+    $page->fillField('title[0][value]', 'OE Showcase Demo Page');
+
+    // Set page description.
+    $page->fillField('field_description[0][value]', 'Page demo description');
+
+    // Add Rich text paragraph.
+    $page->pressButton('Add Rich text');
+    $page->fillField(
+      'field_body[0][subform][field_oe_title][0][value]',
+      'Rich Text paragraph title'
+    );
+    $page->fillField(
+      'field_body[0][subform][field_oe_text_long][0][value]',
+      'Rich Text paragraph Body'
+    );
+
+    // Add a Banner paragraph.
+    $page->pressButton('Add Banner');
+    $page->selectFieldOption(
+      'field_body[1][subform][field_oe_banner_type]',
+      'page_center'
+    );
+    $page->fillField(
+      'field_body[1][subform][field_oe_title][0][value]',
+      'Banner 0 item title'
+    );
+    $page->fillField(
+      'field_body[1][subform][field_oe_text][0][value]',
+      'Banner 0 item Body'
+    );
+
+    // Add Quote Paragraph.
+    $page->pressButton('Add Quote');
+    $page->fillField(
+      'field_body[2][subform][field_oe_plain_text_long][0][value]',
+      'Every moment is a fresh beginning.'
+    );
+    $page->fillField(
+      'field_body[2][subform][field_oe_text][0][value]',
+      'T.S Eliot'
+    );
+
+    // Add a listing item.
+    $page->pressButton('Add Listing item');
+    $page->fillField(
+      'field_body[3][subform][field_oe_link][0][uri]',
+      '<front>'
+    );
+
+    $page->fillField(
+      'field_body[3][subform][field_oe_title][0][value]',
+      'Home Page'
+    );
+
+    $page->fillField(
+      'field_body[3][subform][field_oe_text_long][0][value]',
+      'Listing item description'
+    );
+
+    // Save node.
+    $page->pressButton('Save');
+
+    // Assert node view.
+    $language = \Drupal::languageManager()->getCurrentLanguage()->getId();
+    $assert_session->addressEquals($language . '/pages/oe-showcase-demo-page');
+
+    // Assert Page title.
+    $assert_session->pageTextContains('OE Showcase Demo Page');
+
+    // Assert Description.
+    $assert_session->pageTextContains('Page demo description');
+
+    // Assert Rich Text Title.
+    $assert_session->pageTextContains('Rich Text paragraph title');
+
+    // Assert Rich Text Body.
+    $assert_session->pageTextContains('Rich Text paragraph Body');
+
+    // Assert Quote.
+    $assert_session->pageTextContains('Every moment is a fresh beginning.');
+    $assert_session->pageTextContains('T.S Eliot');
+
+    // Assert Banner.
+    $assert_session->pageTextContains('Banner 0 item title');
+    $assert_session->pageTextContains('Banner 0 item Body');
+
+    // Assert Listing item.
+    $assert_session->pageTextContains('Home');
+    $assert_session->pageTextContains('Listing item description');
+  }
+
+}


composer.json
```diff
@@ -23,7 +23,7 @@
         "openeuropa/oe_media": "^1.13",
         "openeuropa/oe_multilingual": "^1.8",
         "openeuropa/oe_paragraphs": "dev-master",
-        "openeuropa/oe_whitelabel": "1.x-dev"
+        "openeuropa/oe_whitelabel": "1.0.0-alpha1"
```
drishu:
```suggestion
        "openeuropa/oe_whitelabel": "1.0.0-alpha2"
```

composer.json
```diff
@@ -74,12 +74,23 @@
         "composer-exit-on-patch-failure": true,
         "enable-patching": true,
         "installer-types": ["npm-asset", "bower-asset"],
+        "patches": {
+            "drupal/core": {
```
drishu: the issue seems to be in ctools, the newer version of that has the fix, so maybe try to upgrade ctools instead of patching

tests/src/Functional/ShowcasePageTest.php
```diff
@@ -0,0 +1,292 @@
...
+use Drupal\Tests\oe_showcase\Traits\AuthenticationTrait;
+use Drupal\Tests\node\Traits\ContentTypeCreationTrait;
+use Symfony\Component\DomCrawler\Crawler;
+
+/**
+ * Ensure OE Showcase Pages are created.
+ *
+ * @group oe_showcase
+ */
+class ShowcasePageTest extends BrowserTestBase {
```
drishu: no need for the showcase as there wont be any other page tests then this
```suggestion
class PageTest extends BrowserTestBase {
```

tests/src/Functional/ShowcasePageTest.php
```diff
@@ -0,0 +1,292 @@
...
+namespace Drupal\Tests\oe_showcase\Functional;
+
+use Drupal\Tests\BrowserTestBase;
+use Drupal\node\Entity\Node;
+use Drupal\Tests\oe_showcase\Traits\AuthenticationTrait;
+use Drupal\Tests\node\Traits\ContentTypeCreationTrait;
+use Symfony\Component\DomCrawler\Crawler;
+
+/**
+ * Ensure OE Showcase Pages are created.
```
drishu: Lets be more descriptive as to what is going on in this test
```suggestion
 * Page content creation.
```

tests/src/Functional/ShowcasePageTest.php
```diff
@@ -0,0 +1,292 @@
...
+  use ContentTypeCreationTrait;
+
+  /**
+   * {@inheritdoc}
+   */
+  protected static $modules = [
+    'node',
+    'system',
+    'block',
+    'ctools',
```
drishu: you don't need to specify modules that are dependencies of other modules

tests/src/Functional/ShowcasePageTest.php
```diff
@@ -0,0 +1,292 @@
...
+   * {@inheritdoc}
+   */
+  protected $defaultTheme = 'oe_whitelabel';
+
+  /**
+   * A user with permission to administer site configuration.
+   *
+   * @var \Drupal\user\UserInterface
+   */
+  protected $adminUser;
```
drishu: what is the reasoning behind putting the user entity into a class attribute ?

tests/src/Functional/ShowcasePageTest.php
```diff
@@ -0,0 +1,292 @@
...
+
+  /**
+   * The test user permissions.
+   *
+   * @var array
+   */
+  protected $userPermissions = [
+    'access toolbar',
```
drishu: you don't need access to the toolbar, pls use only the ones you need

tests/src/Functional/ShowcasePageTest.php
```diff
@@ -0,0 +1,292 @@
...
+  ];
+
+  /**
+   * Set to TRUE to strict check all configuration saved.
+   *
+   * @var bool
+   *
+   * @see \Drupal\Core\Config\Development\ConfigSchemaChecker
+   */
+  protected $strictConfigSchema = FALSE;
```
drishu: This happens because of config conflict, so lets create a ticket where we address this issue and remove this one done

tests/src/Functional/ShowcasePageTest.php
```diff
@@ -0,0 +1,292 @@
...
+
+  /**
+   * The test node.
+   *
+   * @var array
+   */
+  protected $node;
```
drishu: this doesn't need to be a class attribute

tests/src/Functional/ShowcasePageTest.php
```diff
@@ -0,0 +1,292 @@
...
+
+  /**
+   * The test media document.
+   *
+   * @var array
+   */
+  protected $document;
```
drishu: this doesn't need to be a class attribute

tests/src/Traits/AuthenticationTrait.php
```diff
@@ -0,0 +1,31 @@
...
+   * Disables the forced login.
+   */
+  protected function disableForcedLogin(): void {
+    \Drupal::configFactory()->getEditable('cas.settings')->set('forced_login.enabled', FALSE)->save();
+  }
+
+  /**
+   * Enables the forced login.
+   */
+  protected function enableForcedLogin(): void {
```
drishu: lets not write unused code

tests/src/Traits/AuthenticationTrait.php
```diff
@@ -0,0 +1,31 @@
...
+  *
+  * @package Drupal\Tests\oe_showcase\Traits
+  */
+trait AuthenticationTrait {
+
+  /**
+   * Disables the forced login.
+   */
+  protected function disableForcedLogin(): void {
+    \Drupal::configFactory()->getEditable('cas.settings')->set('forced_login.enabled', FALSE)->save();
```
drishu: if we will be using this in future tests we might create a trait but for now, its not even worth a function because its just one line of code, just add it where you use it

tests/src/Functional/ShowcasePageTest.php
```diff
@@ -0,0 +1,292 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function setUp(): void {
+    parent::setUp();
+    // Disable EU Login.
+    $this->disableForcedLogin();
+
+    $this->adminUser = $this->drupalCreateUser($this->userPermissions, NULL, TRUE);
```
drishu:
```suggestion
  $this->drupalLogin($this->drupalCreateUser([
    'create oe_showcase_page content',
    ..
  ], NULL, TRUE));
```

tests/src/Functional/ShowcasePageTest.php
```diff
@@ -0,0 +1,292 @@
...
+    $this->drupalLogin($this->adminUser);
+
+    // Create page node.
+    $settings = [
+      'title' => 'Demo showcase page',
+      'type' => 'oe_showcase_page',
+      'uid' => \Drupal::currentUser()->id(),
+    ];
+
+    $node = Node::create($settings);
```
drishu:
```suggestion
    $node = Node::create([
      'title' => 'Demo showcase page',
      'type' => 'oe_showcase_page',
      ..
    ]);
```

tests/src/Functional/ShowcasePageTest.php
```diff
@@ -0,0 +1,292 @@
...
+    $node->save();
+    $this->node = $node;
+
+    // Create document media.
+    $pdf_file = file_save_data(file_get_contents(__DIR__ . '/fixtures/example_1.pdf'), 'public://example_1.pdf');
```
drishu: this bit right here deserves its own function because this code refers to creating assets and is not about asserting page, create a helper function and if it will be used in other tests then we can create a trait or base class.

tests/src/Functional/ShowcasePageTest.php
```diff
@@ -0,0 +1,292 @@
...
+
+  /**
+   * Tests the view exists and has a block display.
+   */
+  public function testCreateShowCasePage() {
+    $assert_session = $this->assertSession();
+    $page = $this->getSession()->getPage();
+
+    // Get node to edit.
+    $this->assertNotEmpty($this->node, 'Node found in database.');
```
drishu: this brings no value, if the node is empty for some reason (I dont see why it would be since we controle that scope) the asserts below would fail.

tests/src/Functional/ShowcasePageTest.php
```diff
@@ -0,0 +1,292 @@
...
+
+    $this->drupalGet('node/' . $this->node->id());
+
+    // Assert Page content.
+    $crawler = new Crawler($this->getSession()->getPage()->getContent());
+    // Assert Page title.
+    $assert_session->pageTextContains('OE Showcase Demo Page');
+
+    // Assert Description.
+    $this->assertEquals(
```
drishu: you can just to pageTextContains instead of all these below, you just assert the content, you don't need the crawler to traverse html structure

tests/src/Functional/ShowcasePageTest.php
```diff
@@ -0,0 +1,292 @@
...
+
+    // Assert document.
+    $this->assertEquals(
+      'example_1.pdf',
+      $crawler->filter('div.document__oe-media-file > span > a')->text()
+    );
+
+    // Assert Accordion.
+    $this->assertCount(1, $crawler->filter('div.accordion'));
+    $crawler->filter('div.accordion > div.accordion-item')->each(
```
drishu: stay away from loops, it makes the test more fragile, often code that should fail wont, because you loop through what is instead of what should be

drishu (changes requested):
After rebase please check if any new paragraphs are available.
Also, the test is failing in drone because you need to improve phpunit.xml.dist (in case it is not already after rebase)

tests/src/Functional/ShowcasePageTest.php
GilNovacomm: I've used the crawler to assure the content is actually where it is supposed to be instead of just confirming they exists.
I've removed, maybe we should use this on oe_whitelabel only?

composer.json
GilNovacomm: I've updated Ctools to the latest DEV, then the file_link module complains of the same issue. There is a patch still underwork (https://www.drupal.org/project/file_link/issues/3147517) to fix it.
Unfortunately the patch is still necessary.

composer.json
```diff
@@ -65,13 +65,19 @@
...
     },
     "extra": {
         "composer-exit-on-patch-failure": true,
         "enable-patching": true,
         "installer-types": ["npm-asset", "bower-asset"],
+        "patches": {
+             "drupal/core": {
+                 "https://www.drupal.org/project/drupal/issues/3139335": "https://www.drupal.org/files/issues/2021-02-05/3139335-4.patch"
+             }
+         },
```
donquixote: Seems like this should be in a separate issue, or not?

donquixote:
For the fixtures: Can we move the pdf to `tests/fixtures/` instead of `tests/src/ExistingSite/fixtures/` ?

composer.json
```diff
@@ -65,13 +65,19 @@
     },
     "autoload-dev": {
         "psr-4": {
-            "Drupal\\Tests\\oe_showcase\\": "./tests/src/"
+            "Drupal\\Tests\\oe_showcase\\": "./tests/src/",
+            "Drupal\\Tests\\oe_showcase_page\\": "./tests/src/"
```
donquixote: This looks wrong to me:
1. Determine if really necessary. If you remove it, what happens?
2. The directory would have to be `modules/oe_showcase_page/tests/src/`, not `tests/src/`.

modules/oe_showcase_page/tests/src/ExistingSite/PageTest.php
```diff
@@ -0,0 +1,195 @@
...
+      'create document media',
+    ];
+
+    $this->user = $this->createUser($permissions);
+  }
+
+  /**
+   * Tests the view exists and has a block display.
+   */
+  public function testCreateShowCasePage() {
```
donquixote, line 37: Seems like a mismatch of method name vs doc description.
Was the description copied from elsewhere?

It seems this method does the following:
- Create a oe_showcase_page node and a document media using an API method.
- Use the edit form to fill in some values.
- Visit the node page to verify that all the modified field values are displayed.

So it seems this is mostly about edit + view, not so much about the creation process.
For the creation process we would instead visit the node/add/oe_showcase_page form and fill in the values there, instead of creating via the API.

modules/oe_showcase_page/tests/src/ExistingSite/PageTest.php
```diff
@@ -0,0 +1,195 @@
...
+  }
+
+  /**
+   * Helper function providing a PDF file for tests.
+   *
+   * @return \Drupal\file\Entity\File
+   *   The pdf document.
+   */
+  public function assetTestPdfFile(): File {
+    $pdf_file = file_save_data(file_get_contents(__DIR__ . '/fixtures/example_1.pdf'), 'public://example_1.pdf');
```
donquixote: `file_save_data()` is marked as deprecated.
We should use an injected service.

modules/oe_showcase_page/tests/src/ExistingSite/PageTest.php
```diff
@@ -0,0 +1,195 @@
...
+    // Assert Page title.
+    $assert_session->pageTextContains('OE Showcase Demo Page');
+
+    // Assert Description.
+    $assert_session->pageTextContains('Page demo description');
+
+    // Assert Rich Text Title.
+    $assert_session->pageTextContains('Rich Text paragraph title');
+
+    // // Assert Rich Text Body.
```
donquixote: Duplicate `//` :)

modules/oe_showcase_page/config/install/field.field.node.oe_showcase_page.field_body.yml
```diff
@@ -13,6 +13,7 @@ dependencies:
     - paragraphs.paragraphs_type.oe_quote
     - paragraphs.paragraphs_type.oe_rich_text
     - paragraphs.paragraphs_type.oe_text_feature_media
+    - paragraphs.paragraphs_type.oe_document
```
donquixote: Strange, why is this part of this pull request?
I thought we only want to add a test?

(same for other config noise)

modules/oe_showcase_page/tests/src/ExistingSite/PageTest.php
```diff
@@ -0,0 +1,195 @@
...
+      'oe_media_file' => [
+        'target_id' => $pdf_file->id(),
+      ],
+    ]);
+    $document->save();
+
+    // Mark test content for deletion.
+    $this->markEntityTypeForCleanup('node');
+    $this->markEntityTypeForCleanup('media');
+    $this->markEntityTypeForCleanup('file');
```
donquixote: Why?

donquixote: Hmm I understand now what this does.
But having it in this position has no effect whatsoever, because all the entities are already created.
The entities _are_ cleaned up after the test, but the reason is not what you think:
- The test user is cleaned up because `UserCreationTrait` calls `$this->markEntityForCleanup($user);` which registers the user for cleanup.
- `node_user_predelete()` deletes all content from this user.
- some other entities are deleted as a side effect, e.g. path aliases.

If we change the node author id to 1 (admin), then the node will remain in the database.

If we want to rely on this `markEntityTypeForCleanup()`, we need to call it at the beginning before creating any entities.

donquixote (changes requested):
Hi! Overall not bad, but I think we need to discuss a bit about what exactly we want to test here.
Is it enough to have just one method with create (via API calls) then edit + view? Or should we have more? Not sure. Perhaps good enough for a start? See also my comment in teams group channel.

tests/src/Functional/ShowcasePageTest.php
donquixote: True.
Also the "create " perms seem to be currently not used in the test, because we create those entities via API calls.

modules/oe_showcase_page/tests/src/ExistingSite/PageTest.php
```diff
@@ -0,0 +1,195 @@
...
+    $assert_session->pageTextContains('Listing item description');
+  }
+
+  /**
+   * Helper function providing a PDF file for tests.
+   *
+   * @return \Drupal\file\Entity\File
+   *   The pdf document.
+   */
+  public function assetTestPdfFile(): File {
```
donquixote: Better method name might be createPdfFileEntity() ?
Also in the comment refer to "file entity" to distinguish from the media entity.

modules/oe_showcase_page/tests/src/ExistingSite/PageTest.php
```diff
@@ -0,0 +1,195 @@
...
+    $pdf_file = file_save_data(file_get_contents(__DIR__ . '/fixtures/example_1.pdf'), 'public://example_1.pdf');
+    $pdf_file->setPermanent();
+    $pdf_file->save();
+
+    return $pdf_file;
```
donquixote: Btw, I noticed that the pdf files are not being cleaned up :)
The build/sites/default/files dir fills up with leftover pdfs.

modules/oe_showcase_page/config/install/field.field.node.oe_showcase_page.field_body.yml
GilNovacomm: Hey Andreas, it seems that the paragraph document config was not kept in previous rebases or during the merge in 1.x. The feature is already merged and there's no other remaining PR for showcase page we'll need to correct this config on this PR.

drishu (dismissed):

donquixote (dismissed):

drishu (approved):

donquixote (approved):

PR #32, opened 2021-10-21
OEL-596: [oe_showcase] news default content

abel-santos-corral:
### Description

Create: default content (11 nodes) of type News, create 1 user (with role Editor) and export them to profile.

### Change log

- Added:
11 nodes
1 user.


modules/oe_showcase_news/oe_showcase_news.install
```diff
@@ -0,0 +1,80 @@
+<?php
```
drishu: remove this file we don't need it

drishu (changes requested):
Don't add so many images, this becomes unmanagable, add a default image to the content type and for now that's it

brummbar:
Closing as outdated.

PR #34, opened 2021-10-26
OEL-702: Create Events default content

GilNovacomm:
### Description

Create default content for OE Showcase Event
### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


modules/oe_showcase_event/oe_showcase_event.info.yml
```diff
@@ -17,6 +17,7 @@ dependencies:
 
 config_devel:
   install:
+    - block.block.pagetitle
```
abel-santos-corral: This has to be moved to the default_content module (this is the last module activated. There has more sense

modules/oe_showcase_page/config/install/block.block.pagetitle.yml
```diff
@@ -21,6 +21,7 @@ visibility:
   'entity_bundle:node':
```
abel-santos-corral: Please, Gil, move this block to the default content, and there it will be used by all the future content types and the ones we have. So then whatever module is created is not messing with this. We concentrate all in default_content module of oe_showcase.

abel-santos-corral (changes requested):

modules/oe_showcase_event/oe_showcase_event.info.yml
GilNovacomm: Hey Abel, how about instead we create a Title block per content type? This way instead of concentrating all in the same block (which could make it hard to update when a feature is uninstalled) we give more flexibility to the user to decide which feature title they would like shown, also when uninstalling that particular feature it's block would also get removed.

abel-santos-corral (approved):
Changes had been done. Ready for QA.

brummbar:
Closing as outdated.

PR #37, opened 2021-10-29
OEL-606: OpenEuropa Event Search

GilNovacomm:
### Description

OpenEuropa Event Search

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


modules/oe_showcase_event/config/install/views.view.showcase_events.yml
```diff
@@ -0,0 +1,183 @@
...
+              autosubmit_hide: true
+              input_required: false
+              allow_secondary: false
+              secondary_label: 'Advanced options'
+              secondary_open: false
+            sort:
+              plugin_id: oe_whitelabel_float_end_sort
+              advanced:
+                combine: true
+                combine_rewrite: "Event date Asc|Date asc\r\nEvent date Desc|Date desc\r\nTitle Asc|A-Z\r\nTitle Desc|Z-A"
```
drishu: its clear that its about event, remove "Event" occurrences from the string

modules/oe_showcase_event/oe_showcase_event.info.yml
```diff
@@ -0,0 +1,60 @@
...
+package: OpenEuropa
+
+dependencies:
+  - rdf_skos:rdf_skos
+  - address:address
+  - link_attributes:link_attributes
+  - drupal:datetime_range
+  - drupal:file
+  - drupal:image
+  - drupal:views
```
drishu: modules like views are enabled by the profile already, skip what is enabled already and order the list alphabetically

modules/oe_showcase_event/oe_showcase_event.install
```diff
@@ -0,0 +1,82 @@
...
+    ];
+    \Drupal::service('rdf_skos.skos_graph_configurator')->addGraphs($graphs);
+  }
+}
+
+/**
+ * Implements hook_uninstall().
+ */
+function oe_showcase_event_uninstall(): void {
+  // Remove content type and its nodes.
```
drishu: remove the uninstall, will never be used

modules/oe_showcase_event/oe_showcase_event.module
```diff
@@ -0,0 +1,52 @@
...
+ * @file
+ * Showcase event module.
+ */
+
+/**
+ * Implements link_attributes_plugin_alter().
+ */
+function oe_showcase_event_link_attributes_plugin_alter(&$plugins): void {
+  // Set a default value for the target attribute.
+  $plugins['target']['default_value'] = '_blank';
```
drishu: remove this, the options need to be exported not hardcoded

modules/oe_showcase_event/oe_showcase_event.module
```diff
@@ -0,0 +1,52 @@
...
+    ],
+  ];
+}
+
+/**
+ * Implements hook_form_alter().
+ */
+function oe_showcase_event_form_alter(&$form, &$form_state, $form_id) {
+  if ($form_id == 'facets_form') {
+    if (isset($form['facets']['country'][0]['country']['#options'])) {
```
drishu: there shouldn't be fixes in this repository, if there is any problem report it in a ticket and lets analyse

modules/oe_showcase_event/templates/node--oe-showcase-event--search-result.html.twig
```diff
@@ -0,0 +1,51 @@
...
+        '#id': 'badge',
+        '#fields': {
+          label: item['#title'],
+          attributes: badge_attr
+        }
+      }]) %}
+    {% endif %}
+  {% endfor %}
+{% endif %}
+{% if content.field_oe_showcase_event_image[0] is defined and content.field_oe_showcase_event_image[0]['#markup'] is not empty %}
```
drishu: there is a new version now:

1. create a PR for updating the epic
2. implement solution from https://github.com/openeuropa/oe_showcase/blob/1.x/modules/oe_showcase_search/templates/node--search-result.html.twig

modules/oe_showcase_event/config/install/facets.facet.city.yml
```diff
@@ -0,0 +1,74 @@
+langcode: en
```
drishu, line 1: the city facet should only appear when the country one is selected

drishu (changes requested):
To many files and nodes, add default values and less nodes, 5 are enough

modules/oe_showcase_event/oe_showcase_event.module
GilNovacomm: I've removed. Note that as I've mentioned before, it is not possible to set a default value (without providing a default URL) for the target and export as configuration, so when a user creates a new event they will have to intentionally set target as blank otherwise the external event link will  navigate them away from the site.

See issue: https://www.drupal.org/project/link_attributes/issues/2973266

modules/oe_showcase_event/config/install/facets.facet.city.yml
GilNovacomm: It is already set. It has the same issue as in dates, the hide feature of facets-form is not working. I've opened a ticket for this and will update on Jira.

modules/oe_showcase_event/config/install/views.view.showcase_events.yml
GilNovacomm: I've removed all the files introduced with Events and since its not possible to set a default value for the file field I've set an existing pdf file on the default contents.
This task was done on OEL-702 since its the ticket that handles default content.

brummbar:
Closing as outdated.

PR #48, opened 2022-01-09
OEL-000: Rework install and update composer.

drishu:
### Description

Rework install and update composer.

### Change log

- Changed:
  - update whitelabel version
  - update modules
  - change info modules to "install" key


Maxfire (dismissed):

abel-santos-corral (commented):
Please consider this at Extend page:
* Move all the OpenEuropa Showcase modules to be moved to a OPENEUROPA SHOWCASE category to differentiate it more clearly
* Put default content title to Upper camelcase
* Homogenize the descriptions
* For Default Content we must reflect all the content types, because there we install also Search and we will add Events and News in a short time.

This has been moved to a ticket: https://citnet.tech.ec.europa.eu/CITnet/jira/projects/OEL/issues/OEL-987?filter=allopenissues

abel-santos-corral (approved):
Agreed with the last changes. Green!

Maxfire (approved):

PR #57, opened 2022-02-07
OEL-479: Theme user account form

drishu:
### Description

Theme user account form.

### Change log

- Added:
  - field groups
  - patch to remove table theming
  - arranged fields


modules/oe_showcase_user_profile/src/EventSubscriber/EuLoginAttributesToUserFieldsSubscriber.php
```diff
@@ -36,7 +36,6 @@ class EuLoginAttributesToUserFieldsSubscriber implements EventSubscriberInterfac
    */
   public static function getSubscribedEvents() {
     return [
-      CasHelper::EVENT_POST_LOGIN => 'updateUserData',
       CasHelper::EVENT_PRE_REGISTER => 'processUserProperties',
```
donquixote: Now we have a leftover function `updateUserData()` which is never called.
We can clean this up or we can leave it for now..

modules/oe_showcase_user_profile/tests/src/ExistingSite/IntegrationTest.php
```diff
@@ -21,27 +21,18 @@ public function testIntegration(): void {
     $this->markEntityForCleanup($admin);
     $this->drupalLogin($admin);
 
+    /** @var \Drupal\Tests\WebAssert $assert_session */
```
donquixote: No need, PhpStorm knows the type from the `@return` doc on the method we call.

modules/oe_showcase_user_profile/src/EventSubscriber/EuLoginAttributesToUserFieldsSubscriber.php
donquixote: Also, do we still want to sync the email field?

modules/oe_showcase_user_profile/tests/src/ExistingSite/IntegrationTest.php
```diff
@@ -59,20 +50,33 @@ public function testIntegration(): void {
...
     $assert_session->fieldValueEquals('Country', 'BE');
     $assert_session->fieldValueEquals('Nationality', 'France (http://publications.europa.eu/resource/authority/country/FRA)');
-    $assert_session->fieldValueEquals('Working Languages', 'French (http://publications.europa.eu/resource/authority/language/FRA)');
+    $working_language = $assert_session->optionExists('Working Languages', 'French');
+    $this->assertSame('selected', $working_language->getAttribute('selected'));
+    $page->pressButton('Save');
 
     $this->drupalGet("user/{$user->id()}");
+    $assert_session->pageTextContains('John Doe');
     $assert_session->pageTextContains('User description bio.');
```
donquixote, line 62: @drishu Let's also test user delete, now we have this button enabled?
We can do it directly here: Go to /edit again, click delete button, then delete.
We could also do a new method, but then we'd have to repeat a lot.

modules/oe_showcase_user_profile/config/install/core.entity_form_display.user.user.default.yml
```diff
@@ -103,37 +140,22 @@ content:
...
-  path:
-    type: path
-    weight: 3
-    region: content
-    settings: {  }
-    third_party_settings: {  }
-  timezone:
-    weight: 2
-    region: content
-    settings: {  }
```
donquixote, line 136: @drishu What's with all these settings in the old widget? Are we sure we need none of this in the new widget?

drishu: path timezone so on are now hidden, as for the working languages its a select now, no need for the autocomplete settings.

modules/oe_showcase_user_profile/src/EventSubscriber/EuLoginAttributesToUserFieldsSubscriber.php
drishu: Yes we want to sync mail, I think I'll put the updateUserData back and mark the fields as readonly through a form alter, then we see with PO

donquixote (approved):

Maxfire (approved):

PR #58, opened 2022-02-07
OEL-1098: Theme the user profile page

donquixote:
[OEL-1098](https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-1098)

### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,6 +26,7 @@
         "drupal/purge_users": "dev-3.x",
         "drupal/search_api": "^1.21",
         "drupal/search_api_autocomplete": "^1.5",
+        "drupal/twig_field_value": "^2.0",
         "drupal/user_field_anonymize": "1.x-dev",
         "drupal/user_fields_visibility": "1.x-dev",
         "easyrdf/easyrdf": "1.0.0 as 0.9.1",
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,6 @@
         "drupal/purge_users": "dev-3.x",
         "drupal/search_api": "^1.21",
         "drupal/search_api_autocomplete": "^1.5",
-        "drupal/twig_field_value": "^2.0",
         "drupal/user_field_anonymize": "1.x-dev",
         "drupal/user_fields_visibility": "1.x-dev",
         "easyrdf/easyrdf": "1.0.0 as 0.9.1",
--- a/config/install/contact.settings.yml
+++ b/config/install/contact.settings.yml
@@ -1,0 +1,1 @@
+user_default_enabled: false
--- a/modules/oe_showcase_user_profile/config/install/core.entity_view_display.user.user.full.yml
+++ b/modules/oe_showcase_user_profile/config/install/core.entity_view_display.user.user.full.yml
@@ -102,11 +102,11 @@ content:
     weight: 5
     region: content
   field_date_of_birth:
-    type: datetime_default
+    type: datetime_custom
     label: above
     settings:
       timezone_override: ''
-      format_type: html_date
+      date_format: 'd F Y'
     third_party_settings: {  }
     weight: 5
     region: content
--- a/modules/oe_showcase_user_profile/config/install/core.entity_view_display.user.user.full.yml
+++ b/modules/oe_showcase_user_profile/config/install/core.entity_view_display.user.user.full.yml
@@ -1,0 +1,173 @@
+langcode: en
+status: true
+dependencies:
+  config:
+    - core.entity_view_mode.user.full
+    - field.field.user.user.field_bio
+    - field.field.user.user.field_city_country
+    - field.field.user.user.field_current_position
+    - field.field.user.user.field_date_of_birth
+    - field.field.user.user.field_first_name
+    - field.field.user.user.field_gender
+    - field.field.user.user.field_last_name
+    - field.field.user.user.field_nationality
+    - field.field.user.user.field_organization
+    - field.field.user.user.field_profile_image
+    - field.field.user.user.field_working_languages
+    - image.style.medium
+  module:
+    - address
+    - datetime
+    - field_group
+    - image
+    - rdf_skos
+    - text
+    - user
+third_party_settings:
+  field_group:
+    group_personal_information:
+      children:
+        - field_gender
+        - field_nationality
+        - field_date_of_birth
+        - field_city_country
+      label: 'Personal information'
+      parent_name: ''
+      region: content
+      weight: 7
+      format_type: html_element
+      format_settings:
+        classes: ''
+        show_empty_fields: false
+        id: ''
+        element: div
+        show_label: false
+        label_element: h3
+        label_element_classes: ''
+        attributes: ''
+        effect: none
+        speed: fast
+        pattern: ''
+        pattern_mapping: {  }
+        pattern_variant: ''
+    group_professional_information:
+      children:
+        - field_organization
+        - field_current_position
+        - field_working_languages
+      label: 'Professional information'
+      parent_name: ''
+      region: content
+      weight: 8
+      format_type: html_element
+      format_settings:
+        classes: ''
+        show_empty_fields: false
+        id: ''
+        element: div
+        show_label: false
+        label_element: h3
+        label_element_classes: ''
+        attributes: ''
+        effect: none
+        speed: fast
+        pattern: ''
+        pattern_mapping: {  }
+        pattern_variant: ''
+id: user.user.full
+targetEntityType: user
+bundle: user
+mode: full
+content:
+  field_bio:
+    type: text_default
+    label: above
+    settings: {  }
+    third_party_settings: {  }
+    weight: 3
+    region: content
+  field_city_country:
+    type: address_default
+    label: above
+    settings: {  }
+    third_party_settings: {  }
+    weight: 6
+    region: content
+  field_current_position:
+    type: string
+    label: above
+    settings:
+      link_to_entity: false
+    third_party_settings: {  }
+    weight: 5
+    region: content
+  field_date_of_birth:
+    type: datetime_default
+    label: above
+    settings:
+      timezone_override: ''
+      format_type: html_date
+    third_party_settings: {  }
+    weight: 5
+    region: content
+  field_first_name:
+    type: string
+    label: above
+    settings:
+      link_to_entity: false
+    third_party_settings: {  }
+    weight: 1
+    region: content
+  field_gender:
+    type: skos_concept_entity_reference_label
+    label: above
+    settings:
+      link: false
+    third_party_settings: {  }
+    weight: 3
+    region: content
+  field_last_name:
+    type: string
+    label: above
+    settings:
+      link_to_entity: false
+    third_party_settings: {  }
+    weight: 2
+    region: content
+  field_nationality:
+    type: skos_concept_entity_reference_label
+    label: above
+    settings:
+      link: false
+    third_party_settings: {  }
+    weight: 4
+    region: content
+  field_organization:
+    type: string
+    label: above
+    settings:
+      link_to_entity: false
+    third_party_settings: {  }
+    weight: 4
+    region: content
+  field_profile_image:
+    type: image
+    label: above
+    settings:
+      image_link: ''
+      image_style: medium
+    third_party_settings: {  }
+    weight: 0
+    region: content
+  field_working_languages:
+    type: skos_concept_entity_reference_label
+    label: above
+    settings:
+      link: false
+    third_party_settings: {  }
+    weight: 6
+    region: content
+hidden:
+  langcode: true
+  member_for: true
+  search_api_excerpt: true
--- a/modules/oe_showcase_user_profile/oe_showcase_user_profile.module
+++ b/modules/oe_showcase_user_profile/oe_showcase_user_profile.module
@@ -83,9 +83,10 @@ function oe_showcase_user_profile_form_user_form_alter(&$form, FormStateInterfac
 function oe_showcase_user_profile_preprocess_user__full(array &$variables): void {
   /** @var \Drupal\user\UserInterface $user */
   $user = $variables['user'];
-  $image_item = $user->get('field_profile_image')->get(0);
-  if (!$image_item instanceof ImageItem) {
-    // No profile image found.
+  $image_item_list = $user->get('field_profile_image');
+  $image_item = $image_item_list->get(0);
+  if (!$image_item instanceof ImageItem || !$image_item_list->access()) {
+    // No profile image found, or no view access.
     return;
   }
   $variables['pattern_image'] = ImageValueObject::fromStyledImageItem($image_item, 'medium');
--- a/modules/oe_showcase_user_profile/oe_showcase_user_profile.module
+++ b/modules/oe_showcase_user_profile/oe_showcase_user_profile.module
@@ -9,6 +9,8 @@ declare(strict_types = 1);
 
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Session\AccountInterface;
+use Drupal\image\Plugin\Field\FieldType\ImageItem;
+use Drupal\oe_bootstrap_theme\ValueObject\ImageValueObject;
 use Drupal\user\Entity\User;
 
 /**
@@ -23,6 +25,10 @@ function oe_showcase_user_profile_theme_suggestions_user_alter(array &$suggestio
  */
 function oe_showcase_user_profile_theme() {
   return [
+    'user__full' => [
+      'render element' => 'elements',
+      'base hook' => 'user',
+    ],
     'user__teaser' => [
       'render element' => 'elements',
       'base hook' => 'user',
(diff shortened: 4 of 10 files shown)


modules/oe_showcase_user_profile/templates/user--full.html.twig
```diff
@@ -0,0 +1,47 @@
...
+ */
+#}
+{% spaceless %}
+<article{{ attributes }}>
+  {{ pattern('content_banner', {
+    image: pattern_image,
+    title: content.field_first_name|field_value|render ~ ' ' ~ content.field_last_name|field_value|render,
+    meta: [
+      content.group_professional_information.field_organization|field_value,
+      content.group_professional_information.field_current_position|field_value,
```
drishu, line 15: if you agree lets use the field_value filter also in the teaser (now that we have it)

modules/oe_showcase_user_profile/oe_showcase_user_profile.module
```diff
@@ -48,3 +54,20 @@ function oe_showcase_user_profile_user_format_name_alter(string &$name, AccountI
   }
   $name = implode(' ', $username);
 }
+
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function oe_showcase_user_profile_preprocess_user__full(array &$variables): void {
+  // Set an image value object for the header banner pattern.
```
drishu: no need for the comment, lets comment code only if ambiguous

modules/oe_showcase_user_profile/oe_showcase_user_profile.module
```diff
@@ -48,3 +54,20 @@ function oe_showcase_user_profile_user_format_name_alter(string &$name, AccountI
...
+  $image_element = $variables['elements']['field_profile_image'][0] ?? NULL;
```
drishu: Lets rather bail out early to not execute any additional code if not needed.
instead of checking if certain keys in the render array are present it would be useful and more elegant to use the user object that we have available in the $variables and then make use of the field API
```suggestion
  $user = $variables['user'];

  if ($user->get('field_profile_image')->isEmpty()) {
    return;
  }
```

modules/oe_showcase_user_profile/oe_showcase_user_profile.module
```diff
@@ -48,3 +54,20 @@ function oe_showcase_user_profile_user_format_name_alter(string &$name, AccountI
...
+  $variables['pattern_image'] = NULL;
+  if ($image_element !== NULL) {
```
drishu: no need, we check and bail out if not set earlier

modules/oe_showcase_user_profile/oe_showcase_user_profile.module
```diff
@@ -48,3 +54,20 @@ function oe_showcase_user_profile_user_format_name_alter(string &$name, AccountI
...
   $name = implode(' ', $username);
 }
+
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function oe_showcase_user_profile_preprocess_user__full(array &$variables): void {
+  // Set an image value object for the header banner pattern.
+  $image_element = $variables['elements']['field_profile_image'][0] ?? NULL;
+  $variables['pattern_image'] = NULL;
```
drishu: no need, if there is an image then we have will guarantee a value

modules/oe_showcase_user_profile/oe_showcase_user_profile.module
```diff
@@ -48,3 +54,20 @@ function oe_showcase_user_profile_user_format_name_alter(string &$name, AccountI
...
+  if ($image_element !== NULL) {
+    $image_item = $image_element['#item'] ?? NULL;
```
drishu:
```suggestion
    $image_item = $user->get('field_profile_image')->get(0);
```

modules/oe_showcase_user_profile/oe_showcase_user_profile.module
```diff
@@ -48,3 +54,20 @@ function oe_showcase_user_profile_user_format_name_alter(string &$name, AccountI
...
+    if ($image_item instanceof ImageItem) {
```
drishu: bail out again, this will make the code more readable
```suggestion
    if (!$image_item instanceof ImageItem) {
    return;
  }
```

modules/oe_showcase_user_profile/oe_showcase_user_profile.module
```diff
@@ -48,3 +54,20 @@ function oe_showcase_user_profile_user_format_name_alter(string &$name, AccountI
...
+      $variables['pattern_image'] = isset($image_element['#image_style'])
+        ? ImageValueObject::fromStyledImageItem($image_item, $image_element['#image_style'])
+        : ImageValueObject::fromImageItem($image_item);
+    }
+  }
+}
```
drishu, line 100: since the $variables['elements']['field_profile_image'] will not be printed on page we should add the needed cache directly to variables
you can get the file entity
```
  $cacheability = CacheableMetadata::createFromRenderArray($variables['elements']['field_profile_image'][0]);
  $cacheability->applyTo($variables);
```
best after the is empty check I think

composer.json
```diff
@@ -33,7 +35,7 @@
         "openeuropa/oe_media": "^1.14",
         "openeuropa/oe_multilingual": "^1.9",
         "openeuropa/oe_paragraphs": "dev-master",
-        "openeuropa/oe_whitelabel": "0.1.202202071145"
+        "openeuropa/oe_whitelabel": "0.1.202202080830"
```
drishu: for now do as in 1.x, we'll rebase the epic at a later point

drishu (changes requested):

modules/oe_showcase_user_profile/oe_showcase_user_profile.module
donquixote: ok, in that case we should remove the image in field ui

modules/oe_showcase_user_profile/oe_showcase_user_profile.module
donquixote: But `$variables` is not a render array.

donquixote: Actually your code is correct. `$variables` is not a render array, but still this is the supported way of doing it. 

In ThemeManager::render():
```php
      foreach ($info['preprocess functions'] as $preprocessor_function) {
        if (function_exists($preprocessor_function)) {
          $preprocessor_function($variables, $hook, $info);
        }
      }
      // Allow theme preprocess functions to set $variables['#attached'] and
      // $variables['#cache'] and use them like the corresponding element
      // properties on render arrays. In Drupal 8, this is the (only) officially
      // supported method of attaching bubbleable metadata from preprocess
      // functions. Assets attached here should be associated with the template
      // that we are preprocessing variables for.
      $preprocess_bubbleable = [];
      foreach (['#attached', '#cache'] as $key) {
        if (isset($variables[$key])) {
          $preprocess_bubbleable[$key] = $variables[$key];
        }
      }
```

modules/oe_showcase_user_profile/templates/user--full.html.twig
donquixote: I like the idea, but I think we should do it in a separate issue.
We might have to review more things about the teaser, e.g. the usage of the search variant for the card pattern.

modules/oe_showcase_user_profile/templates/user--full.html.twig
```diff
@@ -0,0 +1,45 @@
...
+    ],
+    content: content.field_bio|field_value,
+    attributes: create_attribute().addClass('mb-4'),
```
drishu: Normally this would go into its own block and be printed in the appropriate region right after the main tag, but since we don't invest in that in v1 lets fix the left padding by adding `ps-0`

modules/oe_showcase_user_profile/templates/user--full.html.twig
```diff
@@ -0,0 +1,45 @@
...
+  }) }}
+
+  {% for group_name, group_label in {
+    group_personal_information: 'Personal information',
```
drishu: reusing html blocks should be done via macros not loops. Lets do a macro of the entire section, something like:
```
{% macro description_list_section(group, content, .. %}
  <section class="my-3">
  ..
{% endmacro %}

description_list_section(group1, content,... 
description_list_section(group2, content,... 
```
but name it and define it whatever way you like, the principle is important here

composer.json
```diff
@@ -24,6 +24,7 @@
         "drupal/purge_users": "dev-3.x",
         "drupal/search_api": "^1.21",
         "drupal/search_api_autocomplete": "^1.5",
+        "drupal/twig_field_value": "^2.0",
```
drishu: this is also a dependency in whitelabel and will stay that way, we can skip it;
Also, some block config changed in whitelabel and now the tabs are duplicated, we have to adapt, not just merge code in,.. we need to rename the local_tasks block to `block.block.oe_whitelabel_local_tasks.yml` and move it to the `config/optional` folder.

modules/oe_showcase_user_profile/oe_showcase_user_profile.module
```diff
@@ -69,3 +76,24 @@ function oe_showcase_user_profile_form_user_form_alter(&$form, FormStateInterfac
...
+ * Implements hook_preprocess_HOOK().
+ */
+function oe_showcase_user_profile_preprocess_user__full(array &$variables): void {
+  /** @var \Drupal\user\UserInterface $user */
+  $user = $variables['user'];
+  $image_item = $user->get('field_profile_image')->get(0);
+  if (!$image_item instanceof ImageItem) {
+    // No profile image found.
+    return;
+  }
```
drishu: we have these modules about restricting access to profile data but this is overwriting that, lets add this check
```suggestion
  $field = $user->get('field_profile_image');
  $image_item = $field->get(0);
  if (!$image_item instanceof ImageItem || !$field->access()) {
    // No profile image found.
    return;
  }
```
Also, the anonymous user is denied access by default, so without this check `$variables['elements']['field_profile_image'][0]` will not be set and visiting a users profile with anonymous will throw an error.

drishu (changes requested):

composer.json
donquixote: Arguments are being made that if we use a dependency directly, we should make it explicit.
There are code review tools that complain if e.g. we use a class from a package that is only implicitly required.

However, arguments can also be made for the opposite, if we assume the implicit dependency will always be there.

Personally I don't like to "assume", but I am also not always consistent about it.

So I am ok to remove it, if that avoids discussion.

donquixote: Arguments are being made that if we use a dependency directly, we should make it explicit.
There are code review tools that complain if e.g. we use a class from a package that is only implicitly required.

However, arguments can also be made for the opposite, if we assume the implicit dependency will always be there.

Personally I don't like to "assume", but I am also not always consistent about it.

So I am ok to remove it, if that avoids discussion.

modules/oe_showcase_user_profile/oe_showcase_user_profile.module
donquixote: Then could we check `['field_profile_image'][0]` instead of the `$field->access()` check?
If no access is given, will `['field_profile_image'][0]` be empty?

As a note, I very much dislike that we have to deal with these details in site-specific code.
There are much better ways to do this.

modules/oe_showcase_user_profile/templates/user--full.html.twig
donquixote: Thanks for the hint! I checked the article about macros, and happily find that there is no risk of polluting a global namespace. So, I will do it.

modules/oe_showcase_user_profile/templates/user--full.html.twig
donquixote: This is a bug in bootstrap itself:
https://github.com/twbs/bootstrap/discussions/35807#discussioncomment-2134560
https://github.com/twbs/bootstrap/issues/35282
I posted about this in teams already, in front-end channel.

Fixes already exist in main branch of bootstrap:
https://github.com/twbs/bootstrap/pull/35825 and some prior commits.

We should upgrade bootstrap, or patch it. Then this will go away.
We should also remove the `<div class="row">` around the `{% include ..card.. %}` in `bcl-content-banner.html.twig`. It has no place here.

donquixote: New issue: https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-1226
This will fix the problem.
Let's ignore this here.

drishu: The way we should do it is either:

1. add a fix and a comment saying "@todo: Remove once OEL-1226 is implemented.", or
2. leave a comment saying "Issue will be fixed by OEL-1226"

In this specific content I'd go with 1.

modules/oe_showcase_user_profile/templates/user--full.html.twig
```diff
@@ -0,0 +1,44 @@
+{#
+/**
+ * @file
+ * Theme override for user in 'full' view mode.
+ */
+#}
+{% spaceless %}
```
drishu: use apply spaceless, see:
https://www.drupal.org/docs/develop/coding-standards/twig-coding-standards#whitespace
since we support 9.2 and above now

modules/oe_showcase_user_profile/templates/user--full.html.twig
donquixote: Even with the `px-0` (which is more correct than `ps-0`), this still looks wrong on mobile.
This must be fixed in bcl.
But ok, I am adding the solution 2 for now.

donquixote: Btw @drishu we can't add comments inside an array, only above it.

drishu (changes requested):
Screenshot 2022-03-01 at 10 28 15
we need to rename config/install/block.block.tabs.yml to config/optional/block.block.oe_whitelabel_local_tasks.yml of course using the 'oe_whitelabel_local_tasks' id so we don't have duplicate tabs;

Also we now have the contact forms enabled so
Screenshot 2022-03-01 at 10 28 37
we need to have the 'contact_default_status' disabled

modules/oe_showcase_user_profile/config/install/core.entity_view_display.user.user.full.yml
```diff
@@ -0,0 +1,173 @@
...
+      link_to_entity: false
+    third_party_settings: {  }
+    weight: 5
+    region: content
+  field_date_of_birth:
+    type: datetime_default
+    label: above
+    settings:
+      timezone_override: ''
+      format_type: html_date
```
drishu: lets use the d F Y format, custom for now until OEL-1224 is resolved

donquixote:
> we need to rename config/install/block.block.tabs.yml to config/optional/block.block.oe_whitelabel_local_tasks.yml of course using the 'oe_whitelabel_local_tasks' id so we don't have duplicate tabs;

is `/optional/` the correct place, or rather `/overrides/`?

> Also we now have the contact forms enabled so we need to have the 'contact_default_status' disabled

modules/oe_showcase_user_profile/config/install/core.entity_view_display.user.user.default.yml
```diff
@@ -49,11 +49,11 @@ content:
     weight: 9
     region: content
   field_date_of_birth:
-    type: datetime_default
+    type: datetime_custom
     label: above
     settings:
       timezone_override: ''
-      format_type: html_date
+      date_format: 'd F Y'
```
drishu: why change it here ? this will have no effect on the user profile page, it should have been changed in the file I requested [core.entity_view_display.user.user.full.yml](https://github.com/openeuropa/oe_showcase/pull/58/files#diff-bafdf476599083808722b2695fc2431537893d02488893b4e8a116a7e24c7ae6)

config/install/contact.settings.yml
```diff
@@ -0,0 +1,5 @@
+default_form: feedback
+flood:
+  limit: 5
+  interval: 3600
+user_default_enabled: false
```
drishu, line 1: leave only this one `user_default_enabled: false` we don't need to export the rest of it, leave those to default values

drishu (changes requested):

config/install/contact.settings.yml
donquixote: I had not considered this as an option. Good idea.

modules/oe_showcase_user_profile/config/install/core.entity_view_display.user.user.default.yml
donquixote: Good catch. Sorry.

drishu (dismissed):

drishu (approved):

escuriola (approved):

PR #68, opened 2022-02-21
OEL-1027: Integrate the event content type

GilNovacomm:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -117,7 +117,7 @@
         },
         "patches": {
             "openeuropa/oe_whitelabel": {
-                "latest 1.x": "https://github.com/openeuropa/oe_whitelabel/compare/0.1.202202221259...1.x.diff"
+                "Adds oe_whitelabel_event. To testing only.": "https://github.com/openeuropa/oe_whitelabel/compare/0.1.202202221259...OEL-1026.diff"
             }
         }
     },
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,6 @@
         "drupal/honeypot": "^2.0",
         "drupal/search_api": "^1.21",
         "drupal/search_api_autocomplete": "^1.5",
-        "drupal/twig_field_value": "^2.0",
         "easyrdf/easyrdf": "1.0.0 as 0.9.1",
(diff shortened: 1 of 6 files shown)


config/install/filter.format.rich_text.yml
```diff
@@ -0,0 +1,16 @@
...
+status: true
+dependencies: {  }
+name: 'Rich text'
+format: rich_text
+weight: 0
+filters:
+  filter_html:
+    id: filter_html
+    provider: filter
+    status: false
```
drishu: looks strange

tests/src/ExistingSite/ShowcaseExistingSiteCreateEventTest.php
```diff
@@ -0,0 +1,119 @@
...
+    $assert_session->statusCodeEquals(403);
+
+    // Create editor user.
+    $user = $this->createUser([]);
+    $user->addRole('editor');
+    $user->save();
+    $this->drupalLogin($user);
+
+    // Assert values.
+    // Assert we have enough permissions.
```
drishu: its clear, no need

drishu (changes requested):

composer.json
```diff
@@ -20,6 +20,7 @@
         "drupal/honeypot": "^2.0",
         "drupal/search_api": "^1.21",
         "drupal/search_api_autocomplete": "^1.5",
+        "drupal/twig_field_value": "^2.0",
```
drishu: twig_field_value is in the oe_whitelabel's require section, so it would be picked up by composer and you dont need to add it then here, now since below patch against 1026 contains this it wont get picked up by composer so we can leave it like this for now, but once 1026 is merged we'll have to remove this, and the patch and use a dev release 0.1.202202... so on

drishu (changes requested):

brummbar (approved):
Good for now.

abel-santos-corral (approved):

drishu (approved):

PR #89, opened 2022-03-11
OEL-1292: Fields and post update changes to make oe_content compatible.

Maxfire:
Fields modified, tests adapted as well and post update to remove and update fields.


composer.json
```diff
@@ -40,9 +40,9 @@
         "openeuropa/oe_media": "^1.14",
         "openeuropa/oe_multilingual": "^1.9",
         "openeuropa/oe_paragraphs": "^1.12",
-        "openeuropa/oe_starter_content": "1.x-dev",
+        "openeuropa/oe_starter_content": "dev-OEL-1290",
         "openeuropa/oe_webtools": "^1.14",
-        "openeuropa/oe_whitelabel": "^1.0.0-alpha5"
+        "openeuropa/oe_whitelabel": "dev-OEL-1291"
```
drishu: for the module it's ok, but the convention for the themes is to use a diff, you can change ..1.x for ..OEL-1291.diff, if it was discussed like this with @brummbar then I dont mind, but I'd keep to our normal way of working

drishu (changes requested):

drishu (dismissed):

modules/oe_showcase_page/config/install/core.entity_view_display.node.oe_showcase_page.teaser.yml
```diff
@@ -19,6 +19,10 @@ content:
     third_party_settings: {  }
     region: content
 hidden:
+  oe_content_content_owner: true
```
drishu, line 22: these will be resolved by OEL-1301

brummbar (changes requested):
All looks good but let's re-export the fields and roles configuration, as they are not up to date.

brummbar (dismissed):

abel-santos-corral (approved):

drishu (approved):

composer.json
```diff
@@ -90,7 +90,7 @@
                 "1.x latest": "https://github.com/openeuropa/oe_bootstrap_theme/compare/1.0.0-alpha7..1.x.diff"
             },
             "openeuropa/oe_whitelabel": {
-                "feature OEL-1291": "https://github.com/openeuropa/oe_whitelabel/compare/1.0.0-alpha5..OEL-1291.diff"
+                "1.x latest": "https://github.com/openeuropa/oe_whitelabel/compare/1.0.0-alpha5..1.x.diff"
```
brummbar, line 93: Let's swap with branch as mac patching is buggy when deleting code.

brummbar (changes requested):
N

PR #90, opened 2022-03-17
OEL-0000: Update composer and fix navigation

drishu:
### Description

Update composer to use latest 1.x version of themes, fix duplicate navigation, fix starter content to use pre-refactoring version.


donquixote (approved):
Looks good.
I confirm that we have a duplicate menu without this fix.

GilNovacomm (approved):

PR #91, opened 2022-03-17
OEL-1167: Theme the multilingual content block

GilNovacomm:
### Description

Require oe_whitelabel_multilingual as a dependency for oe_showcase_navigation.

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


config/optional/block.block.openeuropa_content_language_switcher.yml
```diff
@@ -16,7 +16,7 @@ settings:
   id: oe_multilingual_content_language_switcher
```
drishu: lets remove this file, I think the defaults from whitelabel will be enough

modules/oe_showcase_navigation/oe_showcase_navigation.info.yml
```diff
@@ -6,6 +6,7 @@ package: OpenEuropa Showcase
 dependencies:
   - drupal:views
   - oe_corporate_blocks:oe_corporate_blocks
+  - oe_whitelabel:oe_whitelabel_multilingual
```
drishu: dont we have oe_showcase_multilingual for this ? and add also to oe_showcase.info.yml

drishu (changes requested):

oe_showcase.info.yml
```diff
@@ -51,6 +51,7 @@ install:
   - oe_whitelabel_search
   - oe_whitelabel_starter_news
   - oe_whitelabel_starter_event
+  - oe_whitelabel_multilingual
```
donquixote, line 54: Do we need this, if we already depend on oe_showcase_multilingual?
Or can/should we rely on indirect dependency?

donquixote (commented):
Seems good to me. I left one comment, not sure if we should change this.

oe_showcase.info.yml
drishu: It's the same model as in ewcms, see https://github.com/openeuropa/oe_theme/blob/3.x/modules/oe_theme_content_event/oe_theme_content_event.info.yml and 
https://github.com/ec-europa/ewcms/blob/develop/ewcms.info.yml#L111

but yes, I should take time to ask as to the reason why this practice is being done

donquixote: For module info files and composer.json, if we want to do it right, we should distinguish explicit dependencies vs implicit dependencies.
For a profile, we could ask this question in various ways, but in the end i don't care enough :)
Let's keep it as it is in this PR.

donquixote (dismissed):

drishu (approved):

donquixote (approved):

PR #92, opened 2022-03-18
OEL-1322: PHP 8.0 upgrade for oe_showcase.

escuriola:
### Description

[Insert description here]

### Change log

- Added:
- Changed: PHP version to 8.0
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -78,4 +78,4 @@ pipeline:
 
 matrix:
   PHP_VERSION:
-    - 7.4
+    - 8.0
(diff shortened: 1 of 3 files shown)


brummbar (changes requested):
I made a mistake in the ticket. oe_showcase should work on one php version only. Let's put 8.0.

brummbar (dismissed):

donquixote:
Interesting that we don't need to add token and prophecy-phpunit.
But no need to speculate why - tests are green, we are good.

donquixote (approved):

brummbar (approved):

PR #97, opened 2022-04-01
OEL-1300: Integrate the extended project content type.

Maxfire:


drishu (commented):
OK for now, two considerations:
1. the fields might need to be revisited after merge of the whitelabel PR
2. for future I will create a ticket to move these content type related config into a suibmodule, but not now because it should be done right and include news and event

drishu (changes requested):
One thing I realized we should add is alias pattern like here https://github.com/openeuropa/oe_showcase/blob/1.x/modules/oe_showcase_page/config/install/pathauto.pattern.page_url_alias_pattern.yml

drishu (changes requested):
1) Please double check that anonymous users can see project nodes and skos.
2) Please add a test called **ProjectTest.php** in  tests/src/ExistingSite that:

- asserts users without permission dont have access to the project creation page
- set up assets programatically (images or whatever u need to have ready)
- with a user that has permissions create a project **through the UI**
- save, go to the node url and assert the values are present

tests/src/ExistingSite/ProjectTest.php
```diff
@@ -0,0 +1,150 @@
...
+    $page->fillField('Objective', 'Text Objective');
+    $page->fillField('Impacts', 'Text Impacts');
+    $page->fillField('Achievements and milestones', 'Text Achievements and milestones');
+
+    // Participants.
+    $page->pressButton('Add new participant');
+    $page->fillField('Name', 'Developer participant name');
+    $page->fillField('Acronym', 'Developer participant acronym');
+    $page->fillField('Country', 'BE');
+    $page->fillField('Contribution to the budget', '19.9');
```
donquixote, line 100: The selectors are ambiguous.

donquixote: Also, are we not forgetting fields?
Lead contributors?

tests/src/ExistingSite/ProjectTest.php
```diff
@@ -0,0 +1,150 @@
...
+    $assert_session->pageTextContains('Objective');
+    $assert_session->pageTextContains('Text Objective');
+    $assert_session->pageTextContains('Impacts');
+    $assert_session->pageTextContains('Text Impacts');
+    $assert_session->pageTextContains('Achievements and milestones');
+    $assert_session->pageTextContains('Text Achievements and milestones');
+
+    // Assert participants.
+    $assert_session->pageTextContains('Participants');
+    $assert_session->pageTextContains('Name');
```
donquixote: We should use some kind of selector.
Not sure what's the best solution here.

donquixote: Also, "Name" alone is not very useful to test.
We want to verify that "Name" appears left of "Developer participant name", and all that within the participants section.

donquixote (changes requested):

drishu (dismissed):
enough for now

abel-santos-corral (dismissed):

drishu (approved):
ok for now

donquixote (approved):

PR #98, opened 2022-04-05
OEL-1159: Align templates with modified search variant of card pattern

donquixote:
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-1159

### DIFF
--- a/modules/oe_showcase_search/templates/node--search-result.html.twig
+++ b/modules/oe_showcase_search/templates/node--search-result.html.twig
@@ -18,7 +18,8 @@
   title: _title,
   text: content.body,
   image: {
-    path: content.field_image[0]['#markup'],
+    path: image.src,
+    alt: image.alt,
   },
   badges: _badges,
   meta: [
--- a/modules/oe_showcase_search/templates/node--search-result.html.twig
+++ b/modules/oe_showcase_search/templates/node--search-result.html.twig
@@ -5,7 +5,7 @@
  */
 #}
 {% set _title %}
-  <a class="text-underline-hover" href="{{ url }}">{{ label }}</a>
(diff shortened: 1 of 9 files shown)


modules/oe_showcase_search/templates/node--search-result.html.twig
```diff
@@ -7,18 +7,22 @@
 {% set _title %}
   <a class="text-underline-hover" href="{{ url }}">{{ label }}</a>
```
escuriola: Not related, but text-underline-hover must be changed to standalone class

modules/oe_showcase_search/templates/node--search-result.html.twig
```diff
@@ -7,18 +7,22 @@
...
 {% endset %}
+{% set _badges = [] %}
+{% for _item in content.field_category|field_value %}
+  {% set _badges = _badges|merge([{
+    label: _item,
```
escuriola, line 13: badges are links, and links need to add the class standalone, if not, you'll see an underline: 
image

escuriola (changes requested):

modules/oe_showcase_search/templates/node--search-result.html.twig
donquixote: In the old version there were no links for these category items.
I am pushing a fix that reproduces that old behavior without links.

modules/oe_showcase_search/templates/node--search-result.html.twig
donquixote: Fixed.

escuriola (dismissed):

drishu (approved):

escuriola (approved):

PR #87, opened 2022-03-09
OEL-486: Add search feature integration test

drishu:
### Description

Add search feature integration test

PR #109, opened 2022-04-18
OEL-1247: Integration listing page.

Maxfire:
Integration listing page on showcase.


donquixote:
The sad thing: We now need a (post?) update hook.
If all the config goes to a new submodule (oe_showcase_list_pages), we only need an update hook that enables the new submodule from oe_showcase.install / oe_showcase.post_update.inc.
The goal is that in a project like OESA the configuration will be applied automatically via install-clone.

(Personally I don't like that we have to do this, but this is now the way of things).

EDIT: I just found that most of the oe_* components do module enabling in post_update hooks, not in update hooks.

donquixote:
This makes me wonder:
Are the facets and search index specific to list pages, or would we use the same index for regular search?
If the facets are specific to list pages, can and should we prefix their names?

tests/src/ExistingSite/ListPagesTest.php
```diff
@@ -0,0 +1,119 @@
...
+   */
+  public function testCreateListPages() {
+    // Mark test content for deletion after the test has finished.
+    $this->markEntityTypeForCleanup('node');
+
+    $assert_session = $this->assertSession();
+    $page = $this->getSession()->getPage();
+
+    // Create some test nodes.
+    $day = 00;
```
donquixote:
```suggestion
    $day = 0;
```

tests/src/ExistingSite/ListPagesTest.php
```diff
@@ -0,0 +1,119 @@
...
+    for ($i = 0; $i < 12; $i++) {
+      $date = new DrupalDateTime('20-10-2020');
+      $values = [
+        'title' => 'News number ' . $i,
+        'type' => 'oe_sc_news',
+        'body' => 'This is content number ' . $i,
+        'language' => 'en',
+        'status' => NodeInterface::PUBLISHED,
+        'oe_publication_date' => [
+          'value' => '2022-04-' . $day++,
```
donquixote: Now I see what you are trying to achieve with the `00` above :) but it won't work as you expect.
https://3v4l.org/uSOmV
https://stackoverflow.com/a/1699980/246724
https://3v4l.org/2lRtG

```suggestion
          'value' => sprintf('2022-04-%02d', $i + 1),
```

tests/src/ExistingSite/ListPagesTest.php
donquixote: Actually we can drop the `$day` variable, see below.

tests/src/ExistingSite/ListPagesTest.php
```diff
@@ -0,0 +1,119 @@
...
+    // Mark test content for deletion after the test has finished.
+    $this->markEntityTypeForCleanup('node');
+
+    $assert_session = $this->assertSession();
+    $page = $this->getSession()->getPage();
+
+    // Create some test nodes.
+    $day = 00;
+    for ($i = 0; $i < 12; $i++) {
+      $date = new DrupalDateTime('20-10-2020');
```
donquixote: Why use d-m-Y here, but Y-m-d below? 
Imo, for code, we should always use Y-m-d, it is the least ambiguous format.
```suggestion
      $date = new DrupalDateTime('2020-10-20');
```

tests/src/ExistingSite/ListPagesTest.php
```diff
@@ -0,0 +1,119 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function setUp() {
+    parent::setUp();
+
+    // Create admin user.
+    $role = Role::load($this->createRole([]));
+    $role->setIsAdmin(TRUE);
```
donquixote: Do we need to be admin? Is there currently no other role that can add list pages?
Perhaps there should be, so that people can test it without being admin.

tests/src/ExistingSite/ListPagesTest.php
donquixote: Btw why do we need to set a created timestamp?

tests/src/ExistingSite/ListPagesTest.php
```diff
@@ -0,0 +1,119 @@
...
+      $date = new DrupalDateTime('20-10-2020');
+      $values = [
+        'title' => 'News number ' . $i,
+        'type' => 'oe_sc_news',
+        'body' => 'This is content number ' . $i,
+        'language' => 'en',
+        'status' => NodeInterface::PUBLISHED,
+        'oe_publication_date' => [
+          'value' => '2022-04-' . $day++,
+        ],
```
donquixote: Will it work if we pass the value directly without an array level?
(I don't know, try it)
```suggestion
        'oe_publication_date' => sprintf('2022-04-%02d', $i + 1),
```

tests/src/ExistingSite/ListPagesTest.php
```diff
@@ -0,0 +1,119 @@
...
+    // Assert search.
+    $page->fillField('Publication date', 'gt');
+    $page->fillField('Date', '2022-04-04');
+    $page->pressButton('Search');
+    $assert_session->elementContains('css', 'h4.mb-0', '(7)');
+    $page->fillField('Title', 'News number 8');
+    $page->pressButton('Search');
+    $assert_session->elementContains('css', 'h4.mb-0', '(1)');
+    $assert_session->elementTextEquals('css', 'span.badge.bg-light', 'News number 8');
+  }
```
donquixote: The entire test seems to be mostly a copy of the ListPagesTest in oe_whitelabel.
With some modifications for the new facet.
It feels wrong to me, especially with all the detailed assertions about html classes and attributes.
We mostly want to verify that the facets and the search results show up, and that blocks are not placed in the wrong region. We don't need to test every visual detail.
Also I already don't like the test in whitelabel very much tbh - but this is merged.

A nice example for a search page test is here, https://github.com/openeuropa/oe_showcase/pull/87/files
However, I don't know how far we want to go for something that is already tested in whitelabel.

The copy + paste is definitely a cheap and quick solution for now, and does the job. It can be ok if we plan to refactor it later.

donquixote: We could even have a ListPageAssertion, like we do for patterns :) but that's a wild idea, let's not do it now.

tests/src/ExistingSite/ListPagesTest.php
```diff
@@ -0,0 +1,119 @@
...
+    $assert_session->elementTextEquals('css', 'button[data-bs-toggle="offcanvas"]', 'Filters');
+    // Assert Filters.
+    $assert_session->elementExists('css', 'input[name="oe_sc_news_title"]');
+    $assert_session->elementExists('css', 'select[name="publication_date_op"]');
+    $assert_session->elementExists('css', 'input[name="publication_date_first_date_wrapper[publication_date_first_date][date]"]');
+    $assert_session->elementExists('css', 'input[data-drupal-selector="edit-submit"]');
+    $assert_session->elementExists('css', 'input[data-drupal-selector="edit-reset"]');
+
+    // Assert right column.
+    $assert_session->elementExists('css', 'div.row > div.col-12.col-lg-9.col-xxl-8');
```
donquixote: I don't like the selector, and I don't think we need to verify that a "right column" exists.
I know this is copied from the whitelabel test, but still..

donquixote:
Btw, I notice that the filter icon does not work, but this will be fixed in next version of BCL.

tests/src/ExistingSite/ListPagesTest.php
Maxfire: There's no other role to allow that, nothing specified in the requirements, so we keep it as is.

GilNovacomm (dismissed):

modules/oe_showcase_list_pages/tests/src/ExistingSite/ListPagesTest.php
```diff
@@ -0,0 +1,165 @@
...
+
+declare(strict_types = 1);
+
+use Drupal\Tests\oe_showcase\ExistingSite\ShowcaseExistingSiteTestBase;
+
+use Behat\Mink\Element\NodeElement;
+use Drupal\node\Entity\Node;
+use Drupal\node\NodeInterface;
+use Drupal\Tests\search_api\Functional\ExampleContentTrait;
+use Drupal\user\Entity\Role;
```
donquixote: Please reorder the imports and remove blank line. PhpStorm: Ctrl+Alt+O.
Also please add a namespace declaration.
```suggestion
namespace Drupal\Tests\oe_showcase_list_pages\ExistingSite;

use Behat\Mink\Element\NodeElement;
use Drupal\node\Entity\Node;
use Drupal\node\NodeInterface;
use Drupal\Tests\oe_showcase\ExistingSite\ShowcaseExistingSiteTestBase;
use Drupal\Tests\search_api\Functional\ExampleContentTrait;
use Drupal\user\Entity\Role;
```

modules/oe_showcase_list_pages/tests/src/ExistingSite/ListPagesTest.php
```diff
@@ -0,0 +1,165 @@
...
+    $this->drupalGet('node/' . $node->id() . '/edit');
+    $page->checkField('Override default exposed filters');
+    $page->checkField('emr_plugins_oe_list_page[wrapper][exposed_filters][oe_sc_news_title]');
+    $page->checkField('emr_plugins_oe_list_page[wrapper][exposed_filters][publication_date]');
+    $page->pressButton('Save');
+
+    $this->drupalGet('node/' . $node->id());
+
+    // Assert that the filter form exists.
+    $search_form = $assert_session->elementExists('css', '#oe-list-pages-facets-form');
```
donquixote: Can we call this `$filter_form` instead?

modules/oe_showcase_list_pages/tests/src/ExistingSite/ListPagesTest.php
```diff
@@ -0,0 +1,165 @@
...
+    $title_input = $search_form->findField('Title');
+    $publication_date_input = $search_form->findField('Publication date');
+    $search_button = $search_form->find('css', '#edit-submit');
+    $this->assertNotNull($search_button);
+    $this->assertNotNull($title_input);
+    $this->assertNotNull($publication_date_input);
+
+    // Use the site-wide search for filter news created after 4th april 2022.
```
donquixote: I don't think we are using the site-wide search here.
We are using the filter widgets on the list page.
"Site-wide search" would refer to the global search element in the page header.

modules/oe_showcase_list_pages/tests/src/ExistingSite/ListPagesTest.php
```diff
@@ -0,0 +1,165 @@
...
+      'News number 4',
+      'News number 5',
+      'News number 6',
+      'News number 7',
+      'News number 8',
+      'News number 9',
+      'News number 10',
+      'News number 11',
+    ]);
+    // Use the site-wide search for filter news with specific title.
```
donquixote: Same here, this is not the site-wide search.

donquixote:
```suggestion
    // Filter results by title.
```

modules/oe_showcase_list_pages/tests/src/ExistingSite/ListPagesTest.php
donquixote:
```suggestion
    // Filter results by date.
```

modules/oe_showcase_list_pages/tests/src/ExistingSite/ListPagesTest.php
```diff
@@ -0,0 +1,165 @@
...
+    $this->assertSearchResults([
+      'News number 4',
+      'News number 5',
+      'News number 6',
+      'News number 7',
+      'News number 8',
+      'News number 9',
+      'News number 10',
+      'News number 11',
+    ]);
```
donquixote, line 99: Let's insert a blank line here.
```suggestion
    ]);

```

modules/oe_showcase_list_pages/tests/src/ExistingSite/ListPagesTest.php
```diff
@@ -0,0 +1,165 @@
...
+   * Gets the div element above the search results list.
+   *
+   * @return \Behat\Mink\Element\NodeElement
+   *   Element above the search results list.
+   */
+  protected function getSearchTopRegion(): NodeElement {
+    $element = $this->assertSession()
+      ->elementExists('css', 'div.col-xxl-8');
+    return $element;
+  }
```
donquixote: The method does not do what the doc comment and the name suggest.
It selects the entire right/main column, not just the top region.
I suggest to drop this method, as it is only used once.

modules/oe_showcase_list_pages/tests/src/ExistingSite/ListPagesTest.php
```diff
@@ -0,0 +1,165 @@
...
+
+  /**
+   * Asserts the title above the search results.
+   *
+   * @param int $expected_count
+   *   Expected number of results to be reported in the title.
+   */
+  protected function assertSearchResultsTitle(int $expected_count): void {
+    $title = $this->getSearchTopRegion()->find('css', 'h4.mb-0');
```
donquixote:
```suggestion
    $title = $this->assertSession()->find('css', '.col-xxl-8 h4.mb-0');
```
There is no good selector for this unfortunately, so this is the best we can do.
(And this allows to drop the ->getSearchTopRegion() method)

modules/oe_showcase_list_pages/tests/src/ExistingSite/ListPagesTest.php
```diff
@@ -0,0 +1,165 @@
...
+
+use Behat\Mink\Element\NodeElement;
+use Drupal\node\Entity\Node;
+use Drupal\node\NodeInterface;
+use Drupal\Tests\search_api\Functional\ExampleContentTrait;
+use Drupal\user\Entity\Role;
+
+/**
+ * Class to test List pages content type on existing site tests.
+ */
```
donquixote: The comment text is weird.
```suggestion
/**
 * Tests list pages.
 */
```

modules/oe_showcase_list_pages/README.md
```diff
@@ -0,0 +1,37 @@
...
+The scope of this module is to demo the OpenEuropa Library List pages.
+This module uses [Open Europa Listing Page](https://github.com/openeuropa/oe_list_pages) and enables [Open Europa Whitelabel List Pages](https://github.com/openeuropa/oe_whitelabel). This module contains some facets exported and a dedicated index
+and server for the search.
+
+## Setting up bundles and filters
+
+To create list pages enable the oe_list_pages module:
+```
+drush en oe_list_pages
+```
```
donquixote: Really?
In showcase, everything you need will be already enabled by default.
In a custom project, you want to enable `oe_list_pages` and `oe_whitelabel_list_pages`, if you use whitelabel theme.

modules/oe_showcase_list_pages/README.md
```diff
@@ -0,0 +1,37 @@
...
+By default when enabling oe_list_pages the news bundle as well as the title and publication date filters are available.
+
+### Create list pages:
```
donquixote: No colon (":") after hx elements.

modules/oe_showcase_list_pages/README.md
```diff
@@ -0,0 +1,37 @@
...
+* Go to add/oe_list_page.
+* Enter a title to the list page.
+* Click on ```List Page``` link to collapse the available configuration.
+* Select Content on the ```Source entity type```.
+* Select the bundle on ```Source bundle```.
+* Check ```Override default exposed filters``` and select the desired exposed filters you may as well set default values for these filters.
+
+### Add new bundles you should:
```
donquixote: This phrase is weird. Sounds like Yoda talking :)

modules/oe_showcase_list_pages/README.md
```diff
@@ -0,0 +1,37 @@
...
+To create list pages enable the oe_list_pages module:
+```
+drush en oe_list_pages
+```
+By default when enabling oe_list_pages the news bundle as well as the title and publication date filters are available.
+
+### Create list pages:
+* Go to add/oe_list_page.
+* Enter a title to the list page.
+* Click on ```List Page``` link to collapse the available configuration.
```
donquixote: No triple backticks for inline code snippets.
Also this is not really a code snippet, is it?
Just use regular quotes.
And "expand", not "collapse".
```suggestion
* Expand the "List Page" section in the right panel.
```

modules/oe_showcase_list_pages/README.md
```diff
@@ -0,0 +1,37 @@
...
+* Select Content on the ```Source entity type```.
```
donquixote: This is already the default value.

modules/oe_showcase_list_pages/README.md
```diff
@@ -0,0 +1,37 @@
...
+* Select the bundle on ```Source bundle```.
```
donquixote: Should be "News".

modules/oe_showcase_list_pages/README.md
```diff
@@ -0,0 +1,37 @@
...
+* Check ```Override default exposed filters``` and select the desired exposed filters you may as well set default values for these filters.
+
+### Add new bundles you should:
+* Go to ```config/search/search_api```.
```
donquixote:
```suggestion
* Go to `config/search/search-api`.
```

modules/oe_showcase_list_pages/config/install/search_api.server.oe_whitelabel_list_page_search_server_test.yml
```diff
@@ -0,0 +1,13 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - search_api_db
+id: oe_whitelabel_list_page_search_server_test
+name: oe_whitelabel_list_page_search_server_test
```
donquixote: Can we give this a human label?

modules/oe_showcase_list_pages/config/install/search_api.index.oe_whitelabel_list_page_index_test.yml
```diff
@@ -0,0 +1,60 @@
...
+status: true
+dependencies:
+  config:
+    - field.storage.node.oe_publication_date
+    - search_api.server.oe_whitelabel_list_page_search_server_test
+  module:
+    - node
+    - search_api
+id: oe_whitelabel_list_page_index_test
+name: oe_whitelabel_list_page_index_test
```
donquixote: Can we give this a human label?

modules/oe_showcase_list_pages/config/install/search_api.index.oe_whitelabel_list_page_index_test.yml
```diff
@@ -0,0 +1,60 @@
+langcode: en
+status: true
+dependencies:
+  config:
+    - field.storage.node.oe_publication_date
+    - search_api.server.oe_whitelabel_list_page_search_server_test
+  module:
+    - node
+    - search_api
+id: oe_whitelabel_list_page_index_test
```
donquixote: Why "whitelabel"? and why "_test"?

modules/oe_showcase_list_pages/README.md
donquixote: I was going to say most of this stuff should be documented in oe_list_pages, not here. But perhaps it is ok to do it here, following the issue description.

modules/oe_showcase_list_pages/README.md
```diff
@@ -0,0 +1,37 @@
+# OpenEuropa Showcase List Pages
+
+The scope of this module is to demo the OpenEuropa Library List pages.
+This module uses [Open Europa Listing Page](https://github.com/openeuropa/oe_list_pages) and enables [Open Europa Whitelabel List Pages](https://github.com/openeuropa/oe_whitelabel). This module contains some facets exported and a dedicated index
+and server for the search.
+
+## Setting up bundles and filters
```
donquixote: Title does not really describe what the section is about.

donquixote (changes requested):

modules/oe_showcase_list_pages/README.md
Maxfire: depends cause if the user follows the instructions, could be custom.

modules/oe_showcase_list_pages/tests/src/ExistingSite/ListPagesTest.php
Maxfire: $title = $this->assertSession()->elementExists('css', '.col-xxl-8 h4.mb-0');
:)

modules/oe_showcase_list_pages/README.md
donquixote: Yeah, sorry. This was my earlier assumption that the readme is for users of showcase. But it is for users who create their own website.

modules/oe_showcase_list_pages/README.md
```diff
@@ -0,0 +1,32 @@
...
+### Create list pages
+* Go to add/oe_list_page.
+* Enter a title to the list page.
+* Expand the "List Page" section in the right panel.
+* Select the bundle on "Source bundle".
+* Check "Override default exposed filters" and select the desired exposed filters you may as well set default values for these filters.
+
+### To set up new bundles
+
+* Go to `config/search/search_api`.
```
donquixote:
```suggestion
* Go to `config/search/search-api`.
```

modules/oe_showcase_list_pages/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+The scope of this module is to demo the OpenEuropa Library List pages.
+This module uses [Open Europa Listing Page](https://github.com/openeuropa/oe_list_pages) and enables [Open Europa Whitelabel List Pages](https://github.com/openeuropa/oe_whitelabel). This module contains some facets exported and a dedicated index
+and server for the search.
+
+## Create a list page and setting up bundles and filters
+By default enables oe_list_pages, the news bundle as well as the title and publication date filters are available.
+
+### Create list pages
+* Go to add/oe_list_page.
```
donquixote:
```suggestion
* Go to `node/add/oe_list_page`.
```

modules/oe_showcase_list_pages/README.md
```diff
@@ -0,0 +1,32 @@
+# OpenEuropa Showcase List Pages
+
+The scope of this module is to demo the OpenEuropa Library List pages.
+This module uses [Open Europa Listing Page](https://github.com/openeuropa/oe_list_pages) and enables [Open Europa Whitelabel List Pages](https://github.com/openeuropa/oe_whitelabel). This module contains some facets exported and a dedicated index
+and server for the search.
+
+## Create a list page and setting up bundles and filters
+By default enables oe_list_pages, the news bundle as well as the title and publication date filters are available.
```
donquixote: You changed this sentence, but it is still weird

donquixote: We basically have two cases:
- If you are playing with showcase, all you need to do is enable `oe_showcase_list_pages` and create a custom list page for news.
- If you build your own website without showcase, you would enable `oe_list_pages` and `oe_whitelabel_list_pages`, then you would create search index and facets, then you would create your list page, or you would leave this to editors on production.

With the sentence above, I am not sure which case is being addressed here.
(talk to me in chat)

oe_showcase.post_update.php
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * Implements hook_post_update_NAME().
+ *
+ * Install "OpenEuropa showcase listing pages" module.
+ */
+function oe_showcase_post_update_10001(&$sandbox): void {
+  \Drupal::service('module_installer')->install(['oe_showcase_list_pages']);
+  // Invalidate kernel container to make site info service available.
+  \Drupal::service('kernel')->invalidateContainer();
```
donquixote: We should only do this if we really need it.
The comment does not really tell me why, I suspect it is copied from somewhere.

modules/oe_showcase_list_pages/README.md
donquixote: Resolved, but I added a new comment.

modules/oe_showcase_list_pages/README.md
```diff
@@ -0,0 +1,32 @@
...
+* Add new index.
+* Check Content on "Datasources".
+* Select the desired bundle(s).
+* Click on Save.
+* Click on the tab fields to add the fields do be indexed.
+* Click on save.
+
+### Set up new filters
+Filters are facets on search API module. To create new facets you should:
+* Go to "admin/config/search/facets".
```
donquixote: Let's be consistent.
UI labels in quotes, url paths in backticks.
```suggestion
* Go to `admin/config/search/facets`.
```

modules/oe_showcase_list_pages/README.md
donquixote: Not fixed. But I created a new comment, so I will resolve this one.

modules/oe_showcase_list_pages/config/install/search_api.server.oe_list_page_search_server.yml
```diff
@@ -0,0 +1,13 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - search_api_db
+id: oe_list_page_search_server
```
donquixote: I don't know if `oe_list_page_` is the best prefix. But I say it is good enough for showcase.
(the module name would be `oe_list_pages_` with -s. But I don't even know if this should be our prefix. So let's accept your choice.).

modules/oe_showcase_list_pages/README.md
donquixote: Still a bit weird, but let's resolve this one.

modules/oe_showcase_list_pages/README.md
donquixote: If you can think of an easy improvement, go ahead. Otherwise leave it be: I want to change the readme myself anyway. Perhaps in a follow-up issue.

Maxfire: Improved it a bit

donquixote: Still weird but let's address this in a follow-up.

donquixote (approved):

GilNovacomm (approved):

PR #112, opened 2022-04-22
OEL-1458: Activate block Social share module

abel-santos-corral:
### Description

Activate module oe_webtools_social_share and place block Social share.

### Change log

- Added: config/optional/block.block.showcase_socialshare.yml
- Changed: oe_showcase.info.yml

### DIFF
--- a/config/optional/block.block.showcase_socialshare.yml
+++ b/config/optional/block.block.showcase_socialshare.yml
@@ -1,0 +1,24 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - oe_webtools_social_share
+    - system
+  theme:
+    - oe_whitelabel
+id: showcase_socialshare
+theme: oe_whitelabel
+region: content
+weight: 0
+provider: null
+plugin: social_share
+settings:
+  id: social_share
+  label: 'Social Share'
+  label_display: '0'
+  provider: oe_webtools_social_share
+visibility:
+  request_path:
+    id: request_path
+    negate: false
+    pages: '<front>'
--- a/config/optional/block.block.showcase_socialshare.yml
+++ b/config/optional/block.block.showcase_socialshare.yml
@@ -2,8 +2,8 @@ langcode: en
 status: true
 dependencies:
   module:
+    - node
     - oe_webtools_social_share
-    - system
   theme:
     - oe_whitelabel
 id: showcase_socialshare
@@ -18,7 +18,12 @@ settings:
   label_display: '0'
   provider: oe_webtools_social_share
 visibility:
-  request_path:
-    id: request_path
+  'entity_bundle:node':
+    id: 'entity_bundle:node'
     negate: false
-    pages: '<front>'
+    context_mapping:
+      node: '@node.node_route_context:node'
+    bundles:
+      oe_sc_event: oe_sc_event
+      oe_sc_news: oe_sc_news
+      oe_showcase_page: oe_showcase_page
--- a/modules/oe_showcase_page/templates/oe-webtools-social-share.html.twig
+++ b/modules/oe_showcase_page/templates/oe-webtools-social-share.html.twig
@@ -1,0 +1,12 @@
+{#
+/**
+ * @file
+ * Override template for webtools social share.
+ *
+ * @see ./modules/contrib/oe_webtools/modules/oe_webtools_social_share/templates/oe_webtools_social_share.html.twig
+ */
+#}
+<div class="bcl-social-media-share">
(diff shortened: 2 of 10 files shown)


config/optional/block.block.showcase_socialshare.yml
```diff
@@ -0,0 +1,24 @@
...
+weight: 0
+provider: null
+plugin: social_share
+settings:
+  id: social_share
+  label: 'Social Share'
+  label_display: '0'
+  provider: oe_webtools_social_share
+visibility:
+  request_path:
```
drishu: please enable this for all content types

oe_showcase.info.yml
```diff
@@ -46,6 +46,7 @@ install:
   - oe_showcase_contact_forms
   - oe_showcase_user_profile
   - oe_webtools
+  - oe_webtools_social_share
```
drishu, line 49: add a post_update hook to enable this module

tests/src/ExistingSite/SmokeTest.php
```diff
@@ -14,7 +14,11 @@ class SmokeTest extends ShowcaseExistingSiteTestBase {
    */
   public function testWebsiteIsUp(): void {
     $this->drupalGet('<front>');
-    $this->assertSession()->statusCodeEquals(200);
+    $assert_session = $this->assertSession();
+    $assert_session->statusCodeEquals(200);
+    // Assert only one Block social share and that is inside the content region.
+    $assert_session->elementsCount('css', 'div#block-showcase-socialshare', 1);
+    $assert_session->elementExists('xpath', '//div[@class="container mt-5"]/*//div[@id="block-showcase-socialshare"]');
```
drishu: this line makes the above one redundant, rather:
-grab the parent container, the main page content, use cores cssSelect() or whatever you like and grab the blocks wrapper
-assert the block title is present
-do this with a helper in ShowcaseExistingSiteTestBase
-check for this in all the content type tests: page, news, event
-remove this code from the smoke test

drishu (changes requested):

tests/src/ExistingSite/SmokeTest.php
```diff
@@ -14,7 +14,8 @@ class SmokeTest extends ShowcaseExistingSiteTestBase {
    */
   public function testWebsiteIsUp(): void {
     $this->drupalGet('<front>');
-    $this->assertSession()->statusCodeEquals(200);
+    $assert_session = $this->assertSession();
```
drishu: revert this change, no need to create a variable just for this

tests/src/ExistingSite/ShowcaseExistingSiteCreateEventTest.php
```diff
@@ -113,6 +113,9 @@ public function testCreateEvent() {
     $assert_session->responseContains('image-test.png');
     $assert_session->responseContains('Starter Image test');
     $assert_session->responseContains('Starter Image test alt');
+
+    // Assert Social Share block.
+    $assert_session->elementExists('xpath', '//div[@class="container mt-5"]/*//div[contains(., "Share this page") and @id="block-showcase-socialshare"]');
```
drishu: checking for the id doesn't help, either check for the json, or just the title.

And in the end, now you have to change it in 3 places, so better with the helper method

drishu (changes requested):

drishu (approved):

drishu (dismissed):

brummbar (changes requested):
Code is good but we need to override the template and use a `h3.fw-bold` as per Figma:
image

This is what we have:
image

brummbar:
Just the title, the widget cannot and shouldn't be themed.

modules/oe_showcase_page/templates/oe-webtools-social-share.html.twig
```diff
@@ -0,0 +1,12 @@
+{#
+/**
+ * @file
+ * Override template for webtools social share.
+ *
+ * @see ./modules/contrib/oe_webtools/modules/oe_webtools_social_share/templates/oe_webtools_social_share.html.twig
```
brummbar: Nit:
```suggestion
 * @see ./modules/contrib/oe_webtools/modules/oe_webtools_social_share/templates/oe-webtools-social-share.html.twig
```

brummbar (changes requested):
A very very very small nit.
For the rest the code is good, but it points out that we really need a child theme in our profile.
I'll open the ticket, but this code can move forward now.

brummbar (approved):

drishu (approved):

PR #114, opened 2022-04-25
OEL-1446: Add cookie consent management

abel-santos-corral:
### Description

We need to have cookie consent management at OE Showcase

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:


tests/src/ExistingSite/ShowcaseExistingSiteTestBase.php
```diff
@@ -50,4 +51,16 @@ public function tearDown() {
     parent::tearDown();
   }
 
+  /**
+   * Asserts Cookie Consent block.
+   */
+  protected function assertCookieConsentBlock(): void {
```
drishu: this is not a block, also this needs to appear in all pages, not just content type, so we need to assert in a few scenarios:
-like the front page, 
-yes a node (this way we cover entity urls), 
-you can do also the node listing (the one from core /node),
and for me that would be enough, maybe also just checking front would suffice, but better to have a bit more coverage.
In any case, create a separate test, ad this helper there no here

tests/src/ExistingSite/ShowcaseExistingSiteTestBase.php
```diff
@@ -50,4 +51,16 @@ public function tearDown() {
...
+    $head_content = $this->cssSelect('head')[0];
```
drishu: since this is not a block we really don't care about its placement, just fetch the page content

tests/src/ExistingSite/ShowcaseExistingSiteTestBase.php
```diff
@@ -50,4 +51,16 @@ public function tearDown() {
...
+    $this->assertSession()->elementExists('css', 'script[type="application/json"]');
```
drishu: y, but since this is a tag in the head section, you can just assert the tag is there
like assertStringContainsString('<script type..', $page->getHtml()) something like this

modules/oe_showcase_page/tests/src/ExistingSite/PageTest.php
```diff
@@ -367,6 +367,8 @@ public function testCreateShowCasePage() {
     $assert_session->pageTextContains('First term description');
     $assert_session->pageTextContains('Second term');
     $assert_session->pageTextContains('Second term description');
+
+    $this->assertCookieConsentBlock();
```
drishu: remove

tests/src/ExistingSite/ShowcaseExistingSiteCreateEventTest.php
```diff
@@ -113,6 +113,8 @@ public function testCreateEvent() {
     $assert_session->responseContains('image-test.png');
     $assert_session->responseContains('Starter Image test');
     $assert_session->responseContains('Starter Image test alt');
+
+    $this->assertCookieConsentBlock();
```
drishu: remove

tests/src/ExistingSite/ShowcaseExistingSiteCreateNewsTest.php
```diff
@@ -90,6 +90,8 @@ public function testCreateNews() {
     $assert_session->responseContains('image-test.png');
     $assert_session->responseContains('Starter Image test');
     $assert_session->responseContains('Starter Image test alt');
+
+    $this->assertCookieConsentBlock();
```
drishu: remove

drishu (changes requested):

tests/src/ExistingSite/ShowcaseExistingSiteCookieConsentTest.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\Tests\oe_showcase\ExistingSite;
+
+/**
+ * Class to test Cookie Consent existing.
+ */
+class ShowcaseExistingSiteCookieConsentTest extends ShowcaseExistingSiteTestBase {
```
drishu: CookieConsentTest

tests/src/ExistingSite/ShowcaseExistingSiteCookieConsentTest.php
```diff
@@ -0,0 +1,43 @@
...
+
+  /**
+   * Asserts Cookie Consent block.
+   *
+   * @param string $page
+   *   A list of pages to check the cookie block exists.
+   *
+   * @dataProvider cookieConsentBlockPagesPathProvider
```
drishu: dont use a provider just call drupalGet a few times and the assert

tests/src/ExistingSite/ShowcaseExistingSiteCookieConsentTest.php
```diff
@@ -0,0 +1,43 @@
...
+
+namespace Drupal\Tests\oe_showcase\ExistingSite;
+
+/**
+ * Class to test Cookie Consent existing.
+ */
+class ShowcaseExistingSiteCookieConsentTest extends ShowcaseExistingSiteTestBase {
+
+  /**
+   * Asserts Cookie Consent block.
```
drishu: not a block

tests/src/ExistingSite/ShowcaseExistingSiteCookieConsentTest.php
```diff
@@ -0,0 +1,43 @@
...
+   *
+   * @param string $page
+   *   A list of pages to check the cookie block exists.
+   *
+   * @dataProvider cookieConsentBlockPagesPathProvider
+   */
+  public function testCookieConsentBlock(string $page): void {
```
drishu: not a block

tests/src/ExistingSite/ShowcaseExistingSiteTestBase.php
```diff
@@ -50,4 +50,13 @@ public function tearDown() {
     parent::tearDown();
   }
 
+  /**
+   * Asserts Cookie Consent block.
+   */
+  protected function assertCookieConsentBlock(): void {
```
drishu: not a block

tests/src/ExistingSite/ShowcaseExistingSiteTestBase.php
```diff
@@ -50,4 +50,13 @@ public function tearDown() {
...
+    $page = $this->getSession()->getPage();
```
drishu: no point in creating a variable for a one time use

tests/src/ExistingSite/ShowcaseExistingSiteCookieConsentTest.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\Tests\oe_showcase\ExistingSite;
+
+/**
+ * Class to test Cookie Consent existing.
```
drishu: at least drop the "existing"

drishu (commented):
Take a closer look at the comments pls

oe_showcase.post_update.php
```diff
@@ -38,3 +38,10 @@ function oe_showcase_post_update_00001(&$sandbox) {
 function oe_showcase_post_update_00002(&$sandbox) {
   \Drupal::service('module_installer')->install(['oe_webtools_social_share']);
 }
+
+/**
+ * Enable OE Webtools Cookie Consent component.
+ */
+function oe_showcase_post_update_00003(&$sandbox) {
```
drishu: make this oe_showcase_post_update_00002, as it has not been released yet

tests/src/ExistingSite/ShowcaseExistingSiteTestBase.php
```diff
@@ -73,4 +73,12 @@ protected function assertSocialShareBlock(): void {
     $this->assertStringContainsString(Json::encode($social_share_config), $main_content->getHtml());
   }
 
+  /**
+   * Asserts Cookie Consent block.
+   */
+  protected function assertCookieConsent(): void {
```
drishu: move this to CookieConsentTest, don't know if we will need these somewhere else later but for now we need it only there

drishu (changes requested):

drishu (approved):

brummbar (approved):

PR #116, opened 2022-04-26
OEL-1459: Activate Feedback block

abel-santos-corral:
### Description

We have been requested to add Feedback block at oe_showcase in all pages.
Users with specific permissions should be able to configure the feedback form ID and status.


tests/src/ExistingSite/ShowcasePageFeedbackFormTest.php
```diff
@@ -0,0 +1,97 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function setUp(): void {
+    parent::setUp();
+
+    $this->disableForcedLogin();
+    $this->backupSimpleConfig('oe_webtools_page_feedback.settings');
+    $this->languageManager = $this->container->get('language_manager');
```
brummbar: What is this needed for?

tests/src/ExistingSite/ShowcasePageFeedbackFormTest.php
```diff
@@ -0,0 +1,97 @@
...
+    $this->drupalGet('<front>');
+    $this->assertSession()->elementExists('xpath', "//script[@type='application/json'][.='" . addcslashes('{"service":"dff","id":1234,"lang":"en"}', '\\\'') . "']");
+    $page->clickLink('English');
+    $page->clickLink('português');
+    $this->assertSession()->elementExists('xpath', "//script[@type='application/json'][.='" . addcslashes('{"service":"dff","id":1234,"lang":"pt"}', '\\\'') . "']");
+    // Assert that Page feedback form is placed before the footer
+    // inside footer region.
+    $page_feedback_selector = '//footer[contains(concat(\' \', normalize-space(@class), \' \'), \' bcl-footer \')]/parent::div[1]/preceding-sibling::div[1]';
+    $page_feedback = $this->getSession()->getPage()->find('xpath', $page_feedback_selector);
+    $this->assertEquals('block-showcase-pagefeedbackform', $page_feedback->getAttribute('id'));
```
brummbar: A small improvement since we have an easier way in our theme:
```
$this->assertSession()->elementExists('css', '#block-showcase-pagefeedbackform + #block-oe-whitelabel-neutral-footer');
```
This gets converted to a similar xpath to the above, but it's much easier to read.

oe_showcase.info.yml
```diff
@@ -46,6 +46,7 @@ install:
   - oe_showcase_contact_forms
   - oe_showcase_user_profile
   - oe_webtools
+  - oe_webtools_page_feedback
```
brummbar, line 51: Down below, we need to add our two new configs to the config_devel entries.

brummbar (changes requested):

src/ConfigImporter.php
```diff
@@ -0,0 +1,126 @@
...
+use Drupal\Core\Config\StorageInterface;
+
+/**
+ * Utility class to import configurations from files.
+ *
+ * @todo Possibly move to a service.
+ *
+ * @internal
+ */
+class ConfigImporter {
```
brummbar: This is a 1 to 1 copy from the bootstrap one, which always worked for us too, now that I checked. What are we doing?

tests/src/ExistingSite/ShowcasePageFeedbackFormTest.php
```diff
@@ -0,0 +1,96 @@
...
+    $this->drupalGet('<front>');
+    $this->assertSession()->elementExists('xpath', "//script[@type='application/json'][.='" . addcslashes('{"service":"dff","id":1234,"lang":"en"}', '\\\'') . "']");
+    $page->clickLink('English');
+    $page->clickLink('português');
+    $this->assertSession()->elementExists('xpath', "//script[@type='application/json'][.='" . addcslashes('{"service":"dff","id":1234,"lang":"pt"}', '\\\'') . "']");
+    // Assert that Page feedback form is placed before the footer
+    // inside footer region.
+    $page_feedback_selector = '//footer[contains(concat(\' \', normalize-space(@class), \' \'), \' bcl-footer \')]/parent::div[1]/preceding-sibling::div[1]';
+    $page_feedback = $this->getSession()->getPage()->find('xpath', $page_feedback_selector);
+    $this->assertSession()->elementExists('css', '#block-showcase-pagefeedbackform + #block-oe-whitelabel-neutral-footer');
```
brummbar: The 3rd line was my suggestion to replace the other two.

oe_showcase.post_update.php
```diff
@@ -8,6 +8,7 @@
 declare(strict_types=1);
 
 use Drupal\block\Entity\Block;
+use Drupal\oe_bootstrap_theme\ConfigImporter;
```
abel-santos-corral, line 11: This has to be refered to the oe_showcase element. Please, change it.

brummbar (changes requested):

oe_showcase.post_update.php
brummbar: No, the whole class is useless, as we just renamed a parameter.

brummbar (dismissed):

brummbar (dismissed):

src/ConfigImporter.php
```diff
@@ -0,0 +1,126 @@
...
+   * @param string $profile
+   *   The profile where the folder is located.
+   * @param string $path
+   *   The relative path to the folder, inside the profile.
+   *
+   * @return \Drupal\Core\Config\StorageInterface
+   *   The config file storage.
+   */
+  protected static function getStorage(string $profile, string $path): StorageInterface {
+    return new FileStorage(\Drupal::service('extension.list.module')->getPath($profile) . $path);
```
abel-santos-corral: @escuriola : My error is that I didn't corrected that, instead of that extension.list.module, it's supposed to be extension.list.profile. The only reason to use this configImporter was to get the config/post_update inside a module of the profile. 
But in on a comment @brummbar states that is not necessary to add a configImporter for the small change we are doing. Nevertheless, you have to check if configImporter must need dual use for profile and modules (at oe_bootstrap_theme) and if so, create a ticket on jira.

brummbar: No @abel-santos-corral , the error was to raise a concern for something that was not needed.
This is in the `extension.list.module` service:
image
We don't need to change anything here, because it already takes care of the profiles folder.
We did already the changes anyway for take into account themes or modules in another ticket.

tests/src/ExistingSite/ShowcasePageFeedbackFormTest.php
```diff
@@ -0,0 +1,94 @@
...
+
+  use AuthenticationTrait;
+  use ConfigurationBackupTrait;
+
+  /**
+   * Language manager.
+   *
+   * @var \Drupal\Core\Language\LanguageManagerInterface
+   */
+  protected $languageManager;
```
drishu: this is not used, please remove

drishu (changes requested):

drishu (dismissed):

tests/src/ExistingSite/ShowcasePageFeedbackFormTest.php
```diff
@@ -0,0 +1,87 @@
...
+namespace Drupal\Tests\oe_showcase\ExistingSite;
+
+use Drupal\Tests\oe_showcase\Traits\AuthenticationTrait;
+use Drupal\Tests\oe_showcase\Traits\ConfigurationBackupTrait;
+use weitzman\DrupalTestTraits\ExistingSiteBase;
+
+/**
+ * Tests the Page feedback form block.
+ */
+class ShowcasePageFeedbackFormTest extends ExistingSiteBase {
```
brummbar: Let's use \Drupal\Tests\oe_showcase\ExistingSite\ShowcaseExistingSiteTestBase, in showcase always use our base tests.

brummbar (changes requested):

tests/src/ExistingSite/ShowcasePageFeedbackFormTest.php
```diff
@@ -0,0 +1,86 @@
...
+    $this->disableForcedLogin();
+    $this->backupSimpleConfig('oe_webtools_page_feedback.settings');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function tearDown(): void {
+    $this->enableForcedLogin();
+    $this->restoreConfiguration();
```
brummbar: The parent does this already, and most of the setup too.

brummbar (changes requested):

brummbar (approved):

drishu (approved):

PR #117, opened 2022-04-29
OEL-1498: User management role and settings

GilNovacomm:
### Description

This will allow privileged users to manage users.
The main goals to achieve are:
  - assign other roles without having to rely on UID 1.
  - unblock users that register through EU Login.
  - harden our permission system.

### Change log

- Added: roleassign
- Changed: user.settings


oe_showcase.info.yml
```diff
@@ -38,33 +37,35 @@ install:
...
   - rdf_skos_language_mapping
+  - roleassign
   - taxonomy
   - toolbar
   - ui_patterns
   - ui_patterns_library
   - ui_patterns_settings
   - user
   - views
+  - block
```
Maxfire, line 68: This should go on top.

src/ConfigImporter.php
```diff
@@ -0,0 +1,126 @@
+<?php
```
Maxfire, line 1: If we're using the oe_bootstrap_theme configImporter I guess we can remove this.

tests/src/ExistingSite/ManageUsersRoleTest.php
```diff
@@ -0,0 +1,93 @@
...
+
+    parent::tearDown();
+  }
+
+  /**
+   * Test the "Manage users" role.
+   */
+  public function testManageUsersRole(): void {
+    $page = $this->getSession()->getPage();
+    // Assert users without the "Manage users" role cannot edit user accounts.
```
Maxfire, line 47:
```suggestion
    $assertions = $this->assertSession();
```

tests/src/ExistingSite/ManageUsersRoleTest.php
```diff
@@ -0,0 +1,93 @@
...
+    // Create test user.
+    $user = $this->createUser();
+    $this->drupalGet('admin/people');
+    $this->assertSession()->pageTextContains('Access denied');
```
Maxfire, line 51:
```suggestion
    $assertions = $this->assertSession();
    $assertions->->getPage();
```

composer.json
```diff
@@ -25,6 +25,7 @@
         "drupal/legal": "^2.0",
         "drupal/paragraphs": "^1.13",
         "drupal/purge_users": "dev-3.x#7a4dd2df",
+        "drupal/roleassign": "1.x-dev@dev",
```
drishu: why not use the beta release ?

oe_showcase.post_update.php
```diff
@@ -38,3 +39,18 @@ function oe_showcase_post_update_00001(&$sandbox) {
 function oe_showcase_post_update_00002(&$sandbox) {
   \Drupal::service('module_installer')->install(['oe_webtools_social_share']);
 }
+
+/**
+ * Import the Manage users role.
```
drishu: we are doing more then the role also 'Create' is a better choice

but also merge this with the one above as oe_showcase_post_update_00002 was not released yet

tests/src/ExistingSite/ManageUsersRoleTest.php
```diff
@@ -0,0 +1,93 @@
...
+
+  /**
+   * Test the "Manage users" role.
+   */
+  public function testManageUsersRole(): void {
+    $page = $this->getSession()->getPage();
+    $assertions = $this->assertSession();
+    // Assert users without the "Manage users" role cannot edit user accounts.
+    // Create test user.
+    $user = $this->createUser();
```
drishu: seems redundant, maybe with an editor would make more sense

tests/src/ExistingSite/ManageUsersRoleTest.php
```diff
@@ -0,0 +1,93 @@
...
+    $user_manager->addRole('manage_users');
+    $user_manager->save();
+    $this->drupalLogin($user_manager);
+    $this->drupalGet('admin/people');
+    $assertions->pageTextContains('People');
+
+    // Assure users with the "Manage users" role can assign
+    // a limited set of roles.
+    $roles = user_role_names(FALSE);
+    $config = \Drupal::config('roleassign.settings');
```
drishu: rather with above logged in user, go to the user creation page, and confirm other roles cannot be assigned (as in roles that should not be present on the page)

tests/src/ExistingSite/ManageUsersRoleTest.php
```diff
@@ -0,0 +1,93 @@
...
+    $assertions->pageTextContains('Access denied');
+    $assertions->linkNotExists('Permissions');
+    $this->drupalGet('admin/people/permissions');
+    $assertions->pageTextContains('Access denied');
+
+    // Assure users with Manage users role cannot change Role assign settings.
+    $assertions->linkNotExists('Role assign');
+    $this->drupalGet('admin/people/roleassign');
+    $assertions->pageTextContains('Access denied');
+  }
```
drishu: we also need to test for the other changes: lock permissions, prevent role assignments during bulk import, new users need to be approved upon registration

drishu (changes requested):

tests/src/ExistingSite/ManageUsersRoleTest.php
GilNovacomm: I need it to not have a role in order to test later that a user with Manage users role can assign this user the Editor role.

oe_showcase.post_update.php
```diff
@@ -33,8 +34,20 @@ function oe_showcase_post_update_00001(&$sandbox) {
...
+  // Enable OE Webtools Social Share component.
   \Drupal::service('module_installer')->install(['oe_webtools_social_share']);
+
+  // Import the Manage users role.
+  \Drupal::service('module_installer')->install(['roleassign']);
+  \Drupal::configFactory()->getEditable('user.settings')
+    ->set('register', 'visitors_admin_approval')
+    ->save();
+  $configs = [
+    'roleassign.settings',
```
drishu: this is not the way to do it, it will throw an error

tests/src/ExistingSite/ManageUsersRoleTest.php
```diff
@@ -0,0 +1,122 @@
...
+
+use Drupal\Tests\oe_showcase\Traits\EntityCleanupTrait;
+use Drupal\user\Entity\Role;
+use Drupal\user\Entity\User;
+use weitzman\DrupalTestTraits\ExistingSiteBase;
+
+/**
+ * Tests the Manage users Role.
+ */
+class ManageUsersRoleTest extends ExistingSiteBase {
```
drishu: I dont get why we dont use ShowcaseExistingSiteTestBase

tests/src/ExistingSite/ManageUsersRoleTest.php
```diff
@@ -0,0 +1,122 @@
...
+
+  use AuthenticationTrait;
```
drishu: remove these

tests/src/ExistingSite/ManageUsersRoleTest.php
```diff
@@ -0,0 +1,122 @@
...
+  use ConfigurationBackupTrait;
+  use EntityCleanupTrait;
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function setUp(): void {
```
drishu: adapt this

tests/src/ExistingSite/ManageUsersRoleTest.php
```diff
@@ -0,0 +1,122 @@
...
+
+    $this->disableForcedLogin();
+    $this->backupSimpleConfig('user.role.manage_users');
+    $this->backupSimpleConfig('user.settings');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function tearDown(): void {
```
drishu: remove this

tests/src/ExistingSite/ManageUsersRoleTest.php
drishu: you'r not even logging the user in, what's the point of visiting 'admin/people' as anonymous user, this needs to be removed

tests/src/ExistingSite/ManageUsersRoleTest.php
```diff
@@ -0,0 +1,122 @@
...
+    $this->assertTrue(in_array('editor', $user->getRoles()));
+
+    // Assure users with Manage users role cannot manage permissions or roles.
+    $assertions->linkNotExists('Roles');
+    $this->drupalGet('admin/people/roles');
+    $assertions->pageTextContains('Access denied');
+    $assertions->linkNotExists('Permissions');
+    $this->drupalGet('admin/people/permissions');
+    $assertions->pageTextContains('Access denied');
+
```
drishu: We need to test a user with the permissions to visit this page and assert that the edit and delete links are missing.

tests/src/ExistingSite/ManageUsersRoleTest.php
```diff
@@ -0,0 +1,122 @@
...
+  use EntityCleanupTrait;
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function setUp(): void {
+    parent::setUp();
+
+    $this->disableForcedLogin();
+    $this->backupSimpleConfig('user.role.manage_users');
```
drishu: where are these changed? I dont see it, please could you point it out

tests/src/ExistingSite/ManageUsersRoleTest.php
```diff
@@ -0,0 +1,122 @@
...
+  public function testManageUsersRole(): void {
+    $page = $this->getSession()->getPage();
+    $assertions = $this->assertSession();
+    // Assert users without the "Manage users" role cannot edit user accounts.
+    // Create test user.
+    $user = $this->createUser();
+    $this->drupalGet('admin/people');
+    $assertions->pageTextContains('Access denied');
+
+    // Assert the People administration page is available only for users with
```
drishu: this comment is wrong, its not "only for users with .."

tests/src/ExistingSite/ManageUsersRoleTest.php
```diff
@@ -0,0 +1,122 @@
...
+    // the 'Manage users' role.
+    $user_manager = $this->createUser();
+    $user_manager->addRole('manage_users');
+    $user_manager->save();
+    $this->drupalLogin($user_manager);
+    $this->drupalGet('admin/people');
+    $assertions->pageTextContains('People');
+
+    // Assure users with the "Manage users" role can assign
```
drishu: this is the bulk import test, the comment is wrong

tests/src/ExistingSite/ManageUsersRoleTest.php
```diff
@@ -0,0 +1,122 @@
...
+    // Assure the manager can assign the Editor role.
+    $this->drupalGet('admin/people');
+    $page->selectFieldOption('edit-action',
+      'user_add_role_action.editor');
+    $page->checkField('edit-user-bulk-form-0');
+    $page->pressButton('Apply to selected items');
+    $assertions->pageTextContains('Add the Editor role to the selected user(s)'
+      . ' was applied to 1 item.');
+    $user = User::load($user->id());
+    $this->assertTrue(in_array('editor', $user->getRoles()));
```
drishu: I dont mind the double check but use assertIsArray()

tests/src/ExistingSite/ManageUsersRoleTest.php
```diff
@@ -0,0 +1,122 @@
...
+    $assertions->elementTextNotContains('css', 'div#edit-roles',
+      'Manage users');
+
+    // Assure the manager can assign the Editor role.
+    $this->drupalGet('admin/people');
+    $page->selectFieldOption('edit-action',
+      'user_add_role_action.editor');
+    $page->checkField('edit-user-bulk-form-0');
+    $page->pressButton('Apply to selected items');
+    $assertions->pageTextContains('Add the Editor role to the selected user(s)'
```
drishu: also assert the manage users role is not present

tests/src/ExistingSite/ManageUsersRoleTest.php
```diff
@@ -0,0 +1,122 @@
...
+    $this->drupalGet('admin/people/permissions');
+    $assertions->pageTextContains('Access denied');
+
+    // Assure users with Manage users role cannot change Role assign settings.
+    $assertions->linkNotExists('Role assign');
+    $this->drupalGet('admin/people/roleassign');
+    $assertions->pageTextContains('Access denied');
+
+    // Assert user bulk operation are not present.
+    $this->drupalGet('admin/people');
```
drishu: didn't we go to admin/people before with the same user logged in? seems to check the same thing, this needs some order

tests/src/ExistingSite/ManageUsersRoleTest.php
```diff
@@ -0,0 +1,122 @@
...
+    $assertions->pageTextContains('Role management is disabled in OE Showcase.'
+      . ' Roles and associated permissions are only changeable by users with'
+      . ' Manage users role.');
+
+    // Assert new user needs to be approved upon registration.
+    // Create a "authenticated" user.
+    $authenticated_user = User::create();
+    $authenticated_user->setUsername('authenticated_user')
+      ->save();
+    $this->assertTrue($authenticated_user->isBlocked());
```
drishu: this gives the same result no matter what setting I use, rather go to /user/register, submit the form, assert the result (even load the user and check the status)

drishu (changes requested):
Please rework the test, its a mess.
Pls go over the points in the description and make a plan for each one

tests/src/ExistingSite/ManageUsersRoleTest.php
```diff
@@ -0,0 +1,122 @@
...
+
+    // Assert user bulk operation are not present.
+    $this->drupalGet('admin/people');
+    $assertions->optionNotExists('Action', 'Add the Manage users role to the'
+      . ' selected user(s)');
+    $assertions->optionNotExists('Action', 'Remove the Manage users role from'
+      . ' the selected user(s)');
+
+    // Assert lock permissions.
+    $user_role = Role::load('manage_users');
```
drishu: rather than doing this I would create a user with permissions and try to add a new permission to a role (can be the manage_users).

tests/src/ExistingSite/AuthorisationTest.php
```diff
@@ -0,0 +1,207 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\Tests\oe_showcase\ExistingSite;
+
+/**
+ * Tests that users can only access pages they are authorized to.
```
drishu, line 10: Most of these comments I would have written differently but I wanted to follow ewcms more closely

drishu, line 8: Most of these comments I would have written differently but I wanted to follow ewcms more closely

tests/src/ExistingSite/AuthorisationTest.php
```diff
@@ -0,0 +1,207 @@
...
+    foreach ($paths as $path) {
+      $this->drupalGet($path);
+      $this->assertSession()->statusCodeEquals(403);
+    }
+  }
+
+  /**
+   * Users with the "Manage users" role can assign a limited set of roles.
+   */
+  public function testManageUsersRoleAssign(): void {
```
drishu, line 79: this was under authorisation but I feel it belongs more under UserManagement

drishu, line 69: this was under authorisation but I feel it belongs more under UserManagement

tests/src/ExistingSite/AuthorisationTest.php
```diff
@@ -0,0 +1,207 @@
...
+
+    foreach ($options as $option) {
+      $assert->fieldNotExists($option);
+    }
+  }
+
+  /**
+   * Assert checkboxes are disabled.
+   */
+  protected function assertCheckboxesDisabled(): void {
```
drishu, line 206: this one I copied over from the test I deleted but was 2 tired to make a separate commit with attribution or cherry pick

drishu, line 197: this one I copied over from the test I deleted but was 2 tired to make a separate commit with attribution or cherry pick

tests/src/ExistingSite/EULoginTest.php
```diff
@@ -0,0 +1,146 @@
...
+   */
+  public function testEuLoginIsEnforced(): void {
+    $paths = [
+      '/user/login',
+      '/eulogin',
+    ];
+
+    foreach ($paths as $path) {
+      $this->drupalGet($path);
+      $this->assertSession()->buttonExists('Login!');
```
drishu, line 67: Login! is there only if eulogin mock is used btw

drishu, line 66: Login! is there only if eulogin mock is used btw

tests/src/ExistingSite/EULoginTest.php
```diff
@@ -0,0 +1,146 @@
...
+  /**
+   * Activates a user given its username.
+   *
+   * @var string $username
+   *   The name of the user to be blocked.
+   */
+  protected function activateUser(string $username): void {
+    $user = $this->getUserByName($username);
+
+    if ($user) {
```
drishu: I feel this should fail if user is empty, is copy pasta, so if it was reviewed and demed ok there then..

drishu, line 129: I feel this should fail if user is empty, is copy pasta, so if it was reviewed and demed ok there then..

tests/src/Traits/UserTrait.php
```diff
@@ -0,0 +1,37 @@
...
+trait UserTrait {
+
+  /**
+   * Gets a user by username.
+   *
+   * @param string $username
+   *   The users machine name.
+   *
+   * @return \Drupal\user\UserInterface
+   *   The user.
```
drishu, line 21: majestic

drishu, line 21: majestic

tests/src/Traits/UserTrait.php
```diff
@@ -0,0 +1,37 @@
...
+   */
+  protected function getUserByName(string $username): UserInterface {
+    $users = \Drupal::entityTypeManager()
+      ->getStorage('user')
+      ->loadByProperties([
+        'name' => $username,
+      ]);
+
+    if (!$users) {
+      throw new \Exception(sprintf('The user with the name %s does not exist.', $username));
```
drishu, line 31: copy pasta, feel like I should replace there with ->assert(..

drishu, line 31: copy pasta, feel like I should replace there with ->assert(..

runner.yml.dist
```diff
@@ -19,7 +19,7 @@ drupal:
     # Index content.
     - "./vendor/bin/drush sapi-i"
     # Enable development related modules.
-    - "./vendor/bin/drush en config_devel -y"
+    - "./vendor/bin/drush en config_devel oe_authentication_eulogin_mock -y"
```
drishu, line 22: in ewcms the service is also started, but I dont see the value of that for us, if we use the actual eulogin then we can also test the user fields in a manual test

drishu, line 22: in ewcms the service is also started, but I dont see the value of that for us, if we use the actual eulogin then we can also test the user fields in a manual test

tests/src/ExistingSite/EULoginTest.php
```diff
@@ -0,0 +1,146 @@
...
+
+    // Redirected back to Drupal.
+    $this->assertIsHomepage();
+  }
+
+  /**
+   * Asserts the current address is the homepage.
+   */
+  protected function assertIsHomepage(): void {
+    $this->assertSession()->addressEquals('/en');
```
drishu, line 119: I feel I could have done this better but I'm 2 tired to judge right now, so I leave that 2 next reviewer

drishu (dismissed):
Comments I left above are more things I point out than feedback, I feel I brought enough improvements, and we need to move on, so I approve.

composer.json
```diff
@@ -12,6 +12,7 @@
         "drupal/allowed_formats": "^1.4",
         "drupal/better_exposed_filters": "^5.0",
         "drupal/block_field": "^1.0@RC",
+        "drupal/cas_mock_server": "^1.4",
```
donquixote: This should probably be in require-dev.

oe_showcase.info.yml
```diff
@@ -8,6 +8,7 @@ install:
   - ckeditor
   - config
   - content_translation
+  - cas_mock_server
```
donquixote: No need to enable by default, or is it?

config/post_updates/00003_manage_users/roleassign.settings.yml
```diff
@@ -0,0 +1,3 @@
+roleassign_roles:
+  editor: editor
+  manage_users: '0'
```
donquixote: ~~This file seems not necessary for post update.~~
I don't see this config being referenced in the post update hook.

config/post_updates/00003_manage_users/user.role.manage_users.yml
```diff
@@ -0,0 +1,18 @@
...
+label: 'Manage users'
+weight: 3
+is_admin: null
+permissions:
+  - 'access administration pages'
+  - 'access toolbar'
+  - 'access user profiles'
+  - 'administer users'
+  - 'assign roles'
+  - 'view the administration theme'
```
donquixote, line 18: ~~Also this one: It is a new role, so no post update needed.~~
Forget it, I was wrong.

oe_showcase.post_update.php
```diff
@@ -39,3 +40,17 @@ function oe_showcase_post_update_00002(&$sandbox) {
...
+ */
+function oe_showcase_post_update_00003(&$sandbox) {
+  \Drupal::service('module_installer')->install(['roleassign']);
+  \Drupal::configFactory()->getEditable('user.settings')
+    ->set('register', 'visitors_admin_approval')
+    ->save();
+  $configs = [
+    'user.role.manage_users',
+  ];
+  ConfigImporter::importMultiple('profile', 'oe_showcase', '/config/post_updates/00003_manage_users', $configs);
```
donquixote, line 68: ~~Not needed, this is a new role.~~

Forget it, I was wrong :)
We need to install this config in post update.

config/post_updates/00003_manage_users/roleassign.settings.yml
donquixote: It will be a pain to keep this up to date.
In most recent 1.x we now have `user.role.configure_page_feedback_form`.

oe_showcase.profile
```diff
@@ -16,3 +19,92 @@ function oe_showcase_install_tasks_alter(&$tasks, $install_state): void {
...
+  $permissions_handler = Drupal::service('user.permissions');
+  $permissions = $permissions_handler->getPermissions();
+
+  $permissions_by_provider = [];
+  foreach ($permissions as $permission_name => $permission) {
+    $permissions_by_provider[$permission['provider']][$permission_name] = $permission;
+  }
+
+  foreach ($role_names as $rid => $name) {
+    foreach ($permissions_by_provider as $provider => $permissions) {
```
donquixote, line 55: We are overwriting the outer scope variable `$permissions`. No harm here, but not good practice.
(I know it is copied but still)

donquixote: Also we are not really using the `$provider` variable.
The code is trying to replicate UserPermissionsForm, but it is not needed here.

oe_showcase.profile
```diff
@@ -16,3 +19,92 @@ function oe_showcase_install_tasks_alter(&$tasks, $install_state): void {
...
+      foreach ($permissions as $perm => $permisssion) {
+        $form['permissions'][$perm][$rid]['#attributes']['disabled'] = TRUE;
```
donquixote, line 57: Let's use `'#disabled' => TRUE` instead of `'#attributes']['disabled' => TRUE`, to prevent client-side tampering.
See https://drupal.stackexchange.com/questions/262646/how-to-make-form-field-read-only#comment393084_262647

oe_showcase.profile
```diff
@@ -16,3 +19,92 @@ function oe_showcase_install_tasks_alter(&$tasks, $install_state): void {
...
+  $user = \Drupal::currentUser();
+  if ($user->id() == 1) {
+    return;
+  }
+
+  // Disable submit button.
+  $form['actions']['submit']['#access'] = FALSE;
+
+  // Add warning message.
+  \Drupal::messenger()->addMessage(t('Role management is disabled in OE Showcase. Roles and associated permissions are only changeable by users with Manage users role.'), MessengerInterface::TYPE_WARNING);
```
donquixote, line 41: I think this message is not fully accurate for our situation.
I see it is coming [from ewcms}(https://github.com/ec-europa/ewcms/blob/develop/modules/ewcms_authorisation/ewcms_authorisation.module#L23), but their setup is a bit different.

I think for us it is like this:
- Roles and permissions are managed by developers by deploying new code.
- Users with 'Manage users' role can assign specific roles to users.
- This form is read-only.

oe_showcase.profile
```diff
@@ -16,3 +19,92 @@ function oe_showcase_install_tasks_alter(&$tasks, $install_state): void {
...
+ */
+function oe_showcase_form_user_admin_permissions_alter(&$form, FormStateInterface $form_state): void {
+  // Don't disable for uid=1.
+  $user = \Drupal::currentUser();
+  if ($user->id() == 1) {
+    return;
+  }
+
+  // Disable submit button.
+  $form['actions']['submit']['#access'] = FALSE;
```
donquixote, line 38: I was a bit curious if this is fully reliable against client-side tampering. What if we insert a client-side button?
Doesn't seem to work. The form is not submitted.
So it _seems_ to be safe.

oe_showcase.profile
```diff
@@ -16,3 +19,92 @@ function oe_showcase_install_tasks_alter(&$tasks, $install_state): void {
...
+  $user = \Drupal::currentUser();
+  if ($user->id() == 1) {
+    return;
+  }
+
+  if (empty($form['entities'])) {
+    return;
+  }
+
+  foreach ($form['entities'] as &$entity) {
```
donquixote, line 79: Weird variable name for a form/render element array.

donquixote: But ok if we want to align with ewcms.

oe_showcase.profile
```diff
@@ -16,3 +19,92 @@ function oe_showcase_install_tasks_alter(&$tasks, $install_state): void {
...
+    if (is_array($entity) && isset($entity['operations']) && is_array($entity['operations'])) {
+
+      if (isset($entity['operations']['data']['#links']['edit'])) {
+        unset($entity['operations']['data']['#links']['edit']);
+      }
+
+      if (isset($entity['operations']['data']['#links']['delete'])) {
+        unset($entity['operations']['data']['#links']['delete']);
+      }
```
donquixote, line 88: So we remove 'edit' and 'delete' but we keep 'Edit permissions' and 'Translate'.
Ok I guess.
The 'Edit permissions' goes to the read-only form so it is fine.

oe_showcase.profile
```diff
@@ -16,3 +19,92 @@ function oe_showcase_install_tasks_alter(&$tasks, $install_state): void {
...
+ * Alter the assignable roles.
+ */
+function oe_showcase_form_bulk_add_cas_users_alter(&$form, FormStateInterface $form_state): void {
+  if (roleassign_restrict_access()) {
+    // Get all roles that are available.
+    $roles = user_role_names(TRUE);
+
+    // Get roles that are available for assignment.
+    $assignable_roles = array_intersect_key($roles, array_filter(\Drupal::config('roleassign.settings')
+      ->get('roleassign_roles')));
```
donquixote, line 105: We could apply the array_intersect_key directly on `$form['roles']['#options']` instead of calling `user_role_names()` first.
But let's keep aligned with ewcms.

donquixote: Also the local variable `$assignable_roles` is a bit pointless if it contains 1:1 what we write to '#options'.
A local var for the rolassign setting would make more sense, as it would help to split up the long expression.

I see this is copied from roleassign module itself, so whatever.

tests/src/ExistingSite/AuthorisationTest.php
```diff
@@ -0,0 +1,207 @@
...
+      '/admin/content',
+      '/admin/people',
+      '/admin/structure',
+      '/node/add',
+    ];
+
+    foreach ($paths as $path) {
+      $this->drupalGet($path);
+      $this->assertSession()->statusCodeEquals(403);
+    }
```
donquixote, line 52: Shouldn't we login first, if we want to test authenticated user? Or does this magically happen somewhere?

tests/src/ExistingSite/AuthorisationTest.php
```diff
@@ -0,0 +1,207 @@
...
+      '/admin/people/roles',
+      '/admin/people/role-settings',
+      '/admin/people/roleassign',
+    ];
+
+    foreach ($paths as $path) {
+      $this->drupalGet($path);
+      $this->assertSession()->statusCodeEquals(403);
+    }
+  }
```
donquixote, line 74: This method is pointless, add the urls to the tests above.

donquixote: (if we are still testing for anon and authenticated)

tests/src/ExistingSite/AuthorisationTest.php
donquixote: I think we should have one method per role we want to test, and cover paths with and without access.
Or we could use a data provider - but let's keep it simple for now.

tests/src/ExistingSite/AuthorisationTest.php
```diff
@@ -0,0 +1,207 @@
...
+
+  /**
+   * Users with the "Manage users" role can assign a limited set of roles.
+   */
+  public function testManageUsersRoleAssign(): void {
+    $user = $this->createUser();
+    $user->addRole('manage_users');
+    $user->save();
+
+    $this->drupalLogin($user);
```
donquixote: We call this regularly, so what about a helper to be called like this:
```suggestion
    $this->loginWithRoles(['manage_users']);
```

tests/src/ExistingSite/UserManagementTest.php
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  /**
+   * Finds the update checkbox in the users row and checks it.
+   *
+   * @param string $username
+   *   The users machine name.
+   */
+  protected function checkUsersUpdateCheckbox(string $username): void {
+    $link = $this->getSession()->getPage()->findLink($username);
```
donquixote, line 48: My IDE (+ EA plugin) would like to see `$this->assertNotNull($link)` here to avoid inspection warning below.
But perhaps we don't care in tests, and use the error as test fail.

tests/src/ExistingSite/AuthorisationTest.php
```diff
@@ -0,0 +1,207 @@
...
+
+    // Test roles availability in the user listing page.
+    $this->drupalGet('/admin/people');
+    $this->assertRoleAvailableInAction(['Editor']);
+    $this->assertRoleNotAvailableInAction(['Manage users']);
+
+    // Test roles availability in the user import page.
+    $this->drupalGet('/admin/people/create/cas-bulk');
+    $this->assertRoleOptionsAvailable(['Editor']);
+    $this->assertRoleOptionsNotAvailable(['Manage users']);
```
donquixote, line 95: I think this part should go into UserManagementTest.
Instead of testing individual operations we can simply check the complete list of bulk update options. This way we test everything the user manager can do.
I think we can't do a negative test, because users without this perm can't even visit the page..

tests/src/ExistingSite/AuthorisationTest.php
```diff
@@ -0,0 +1,207 @@
...
+
+    $this->drupalLogin($user);
+    $this->drupalGet('/admin/people');
+    $this->assertSession()->statusCodeEquals(403);
+
+    $this->drupalGet("user/{$user->id()}/edit");
+    $this->assertRoleOptionsNotAvailable([
+      'Editor',
+      'Manage users',
+    ]);
```
donquixote: This test is a bit pointless:
- We don't even have a role field on user edit form. We disabled it in the form settings. So the user with 'manage_users' role would also not get these options.
- Lack of contrasting test for user 'manage_users' role. -> this would reveal why it's pointless.
- Assertion does not distinguish between not having these specific role checkboxes vs not having a roles input group at all.

tests/src/ExistingSite/AuthorisationTest.php
```diff
@@ -0,0 +1,207 @@
...
+   * Users without the "Manage users" role cannot edit user accounts.
+   */
+  public function testOtherRolesCannotManageUsers(): void {
+    $user = $this->createUser();
+    $user->addRole('editor');
+    $user->save();
+
+    $this->drupalLogin($user);
+    $this->drupalGet('/admin/people');
+    $this->assertSession()->statusCodeEquals(403);
```
donquixote, line 108: This can be covered in the route tests above.

tests/src/ExistingSite/AuthorisationTest.php
donquixote: > I think we can't do a negative test, because users without this perm can't even visit the page..

actually yes we can with custom perm

tests/src/ExistingSite/AuthorisationTest.php
```diff
@@ -0,0 +1,207 @@
...
+    $assert = $this->assertSession();
+
+    $this->drupalLogin($this->createUser(['administer permissions']));
+
+    $this->drupalGet('/admin/people/permissions');
+
+    $assert->pageTextContains('Role management is disabled in OE Showcase. Roles and associated permissions are only changeable by users with Manage users role.');
+    $assert->buttonNotExists('edit-submit');
+    // Assert all checkboxes are disabled on permissions page.
+    $this->assertCheckboxesDisabled();
```
donquixote, line 124: Does this really need to be a helper method?
Seems it is only used 1x.

tests/src/ExistingSite/AuthorisationTest.php
```diff
@@ -0,0 +1,207 @@
...
+    }
+  }
+
+  /**
+   * Assert checkboxes are disabled.
+   */
+  protected function assertCheckboxesDisabled(): void {
+    $checkboxes = $this->getSession()
+      ->getPage()
+      ->findAll('css', '.form-checkbox');
```
donquixote, line 209: Do we want to rely on `'.form-checkbox'`?
If for some reason the class is removed from checkbox elements, then we will get a passing test that checks nothing.

donquixote: Maybe assert a min count? E.g. min 30 checkboxes. And use different kind of selector?
Use count queries and compare count of `input[type=checkbox][disabled]` vs count of `input[type=checkbox]`?
Target the form?

tests/src/ExistingSite/AuthorisationTest.php
```diff
@@ -0,0 +1,207 @@
...
+   * @param array $options
+   *   The list of options.
+   */
+  protected function assertRoleOptionsNotAvailable(array $options): void {
+    $assert = $this->assertSession();
+
+    foreach ($options as $option) {
+      $assert->fieldNotExists($option);
+    }
+  }
```
donquixote, line 201: These two methods could be combined:
Assert the full list of role options.

tests/src/ExistingSite/AuthorisationTest.php
```diff
@@ -0,0 +1,207 @@
...
+   */
+  public function testRolesManagementAccess(): void {
+    $assert = $this->assertSession();
+    $this->drupalLogin($this->createUser(['administer permissions']));
+
+    $this->drupalGet('/admin/people/roles');
+    $assert->statusCodeEquals(200);
+
+    $assert->linkNotExistsExact('Edit');
+    $assert->linkNotExistsExact('Delete');
```
donquixote, line 138: There should be a contrasting test for user with manage_users who _can_ see these links.

donquixote: Also we should verify that the other links are still there, and that the list of roles is not empty.

tests/src/ExistingSite/AuthorisationTest.php
donquixote: Also check that the form is still there.
E.g. we could check one or a few example role/perm checkboxes.

tests/src/ExistingSite/EULoginTest.php
donquixote: And the existence of 'Login!' button tells us reliably that "EU login is enforced"? Interesting.

tests/src/ExistingSite/EULoginTest.php
```diff
@@ -0,0 +1,146 @@
...
+    $this->container->get('cas_mock_server.user_manager')
+      ->addUser([
+        'username' => 'chucknorris',
+        'email' => 'texasranger@chucknorris.com.eu',
+        'password' => 'Qwerty098',
+        'firstName' => 'Chuck',
+        'lastName' => 'Norris',
+        'domain' => 'eu.europa.ec',
+      ]);
+    $this->markEntityTypeForCleanup('user');
```
donquixote, line 40: Shouldn't this be called before creating entities?
Or in case of user it might not be needed at all?

donquixote: How to test: Run the test, see if leftover user is still there.

tests/src/ExistingSite/EULoginTest.php
```diff
@@ -0,0 +1,146 @@
...
+
+    $this->drupalGet('/user/login');
+    $page->fillField('Username or e-mail address', 'texasranger@chucknorris.com.eu');
+    $page->fillField('Password', 'Qwerty098');
+    $page->pressButton('Login!');
+
+    // Redirected back to Drupal and user is blocked.
+    $assert->pageTextContains('Your account is blocked or has not been activated. Please contact a site administrator.');
+    $this->assertUserAccountNotActive('chucknorris');
+    $this->assertIsHomepage();
```
donquixote, line 87: Is there no way to assert if you are currently logged in or not?
I assume the first attempt fails, the second attempt is successful, and then the logout is successful again.

tests/src/ExistingSite/EULoginTest.php
donquixote: Yeah seems a bit awkward, but it does the job.
It does not check query parameters but that's ok.

tests/src/ExistingSite/EULoginTest.php
donquixote: Indeed we should assert not null instead of this if().
I did not find the tests in ewcms, but I believe you saying it is copypasta. But we should fix where it makes sense, and possibly report back to the source after.

runner.yml.dist
donquixote: No opinion on this atm, I would need to try it myself.

donquixote:
One thing I notice:
The roles element is disabled in the used edit form display.
This means the only way to assign roles is via bulk edit.
Is this how we want it to be?

oe_showcase.profile
drishu: dont want to touch it if no adaptation needed

oe_showcase.profile
drishu: dont want to touch

oe_showcase.profile
drishu: dont want to touch

oe_showcase.profile
drishu: ok

oe_showcase.profile
drishu: dont want to touch

oe_showcase.profile
drishu: y

oe_showcase.profile
drishu: yes, lets keep aligned

config/post_updates/00003_manage_users/roleassign.settings.yml
drishu: done

tests/src/ExistingSite/AuthorisationTest.php
drishu: forgot the create the user with the role, thx

tests/src/ExistingSite/AuthorisationTest.php
drishu: so it was like this in ewcms where I copied the approach from, again, I would have done it slightly differently but the scope of the ticket was to have ewcms as a guideline

tests/src/ExistingSite/AuthorisationTest.php
drishu: its a bit gray here, I think the logic is that we assert the access to that functionality is restricted

tests/src/ExistingSite/AuthorisationTest.php
drishu: in ewcms they test more roles, we will also have more roles, so I think this partitioning will start to pay off soon

tests/src/ExistingSite/AuthorisationTest.php
drishu: for readability

tests/src/ExistingSite/AuthorisationTest.php
drishu: we can discuss in a follow up

tests/src/ExistingSite/AuthorisationTest.php
drishu: yes, I will assert count

tests/src/ExistingSite/EULoginTest.php
drishu: yes

tests/src/ExistingSite/EULoginTest.php
drishu: well keep like this

tests/src/ExistingSite/AuthorisationTest.php
drishu: follow up

tests/src/ExistingSite/EULoginTest.php
drishu: I tested, it works fine and its needed, we create a user through eulogin mock so we need to delete him

tests/src/ExistingSite/EULoginTest.php
drishu: its not fancy but I added something for it

tests/src/ExistingSite/AuthorisationTest.php
drishu: > One thing I notice: The roles element is disabled in the used edit form display. This means the only way to assign roles is via bulk edit. Is this how we want it to be?

No, we should create a ticket to not hide those elements via form display but via access in form alter

tests/src/Traits/UserTrait.php
brummbar: No, this is not an assertion, this is an helper method. The test using this can catch and re-launch if needed.

oe_showcase.post_update.php
```diff
@@ -45,3 +45,24 @@ function oe_showcase_post_update_00002(&$sandbox) {
...
+  \Drupal::service('module_installer')->install(['roleassign']);
+  \Drupal::configFactory()->getEditable('user.settings')
+    ->set('register', 'visitors_admin_approval')
+    ->save();
+
+  $roleassign_config = \Drupal::configFactory()->getEditable('roleassign.settings');
+  $roleassign_roles = $roleassign_config->get('roleassign_roles');
+  $roleassign_roles['editor'] = 'editor';
+  $roleassign_roles['manage_users'] = '0';
+  $roleassign_config->set('roleassign_roles', $roleassign_roles)->save();
```
donquixote, line 63: One more role needed.
Otherwise ok for me.

oe_showcase.profile
```diff
@@ -16,3 +19,92 @@ function oe_showcase_install_tasks_alter(&$tasks, $install_state): void {
...
+  }
+
+  foreach ($role_names as $rid => $name) {
+    foreach ($permissions_by_provider as $provider => $permissions) {
+      foreach ($permissions as $perm => $permisssion) {
+        $form['permissions'][$perm][$rid]['#attributes']['disabled'] = TRUE;
+      }
+    }
+  }
+}
```
donquixote, line 61: I've been thinking: What we are doing here is super weird.
All of the modifications on permissions page and roles page only apply to users with `'administer permissions'`.
However, none of the roles we create have this perm.

In ewcms it is a bit different:
- The 'manage_patform_user' role has 'administer permissions', but not 'assign roles'. Don't ask me why!
- The 'manage_users' role has 'assign roles', but not 'administer permissions'.
- The custom form alter hooks make sure that the 'administer (roles and) permissions' is effectively just 'view (roles and) permissions'. But this is only relevant for 'manage_platform_user' role.

donquixote: Anyway no time to change it now.

tests/src/Traits/UserTrait.php
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    return reset($users);
+  }
+
+  /**
+   * Creates a user, assigns given roles and logs the user in.
```
donquixote: Not really, no login happens here.

donquixote:
```suggestion
   * Creates a user with given roles.
```

tests/src/ExistingSite/AuthorisationTest.php
```diff
@@ -0,0 +1,211 @@
...
+   */
+  public function testManageUsersAccess(): void {
+    $user = $this->createUserWithRoles(['manage_users']);
+    $this->drupalLogin($user);
+
+    $paths = [
+      '/admin/people/roles',
+      '/admin/people/roles/add',
+      '/admin/people/role-settings',
+      '/admin/people/roleassign',
```
donquixote, line 64: I think we can add admin/people/permissions

tests/src/ExistingSite/AuthorisationTest.php
```diff
@@ -0,0 +1,211 @@
...
+
+    $this->drupalGet('/admin/people/roles');
+    $assert->statusCodeEquals(200);
+
+    $assert->linkNotExistsExact('Edit');
+    $assert->linkNotExistsExact('Delete');
+
+    $assert->linkExists('Add role');
+    $this->drupalGet('/admin/people/roles/add');
+    $assert->statusCodeEquals(200);
```
donquixote, line 142: This seems like a bug: You can add roles but you cannot edit?
But let's leave it like this, this entire functionality might be pointless anyway if nobody has 'administer permissions'.

tests/src/ExistingSite/EULoginTest.php
donquixote: hm I don't like relying on text not exists.
But let's keep it.

tests/src/Traits/UserTrait.php
donquixote: I don't think the separation is so strict. There are assert calls in other places, e.g. in core UserCreationTrait:.createUser().
And tests usually don't want to catch exceptions, they assume it works, and if not we want the test to fail.
But let's keep it for now. An uncaught non-assert exception is also a kind of fail.

tests/src/ExistingSite/AuthorisationTest.php
donquixote: Readability argument only works if the reader trusts that the helper method will do what is promised :)
But let's keep it
(resolve after reading)

oe_showcase.profile
donquixote: (resolve after reading)

tests/src/ExistingSite/AuthorisationTest.php
donquixote: (resolve after reading)

drishu: yes, the request was to prevent "user manager" specifically to access the role add

donquixote (approved):

brummbar (approved):

PR #126, opened 2022-05-12
OEL-1463: Update pager test assertions for first/last.

donquixote:


modules/oe_showcase_search/tests/src/ExistingSite/IntegrationTest.php
```diff
@@ -225,32 +225,54 @@ protected function assertSearchResults(array $expected): void {
...
+
+    $get_page_index = static function (NodeElement $link): string {
+      $href = $link->getAttribute('href');
+      $query_string = parse_url($href, PHP_URL_QUERY);
+      parse_str($query_string, $query);
+      return $query['page'];
+    };
+
+    if ($active === $total - 1) {
+      $active_link = $pagination->find('css', 'li.active:last-child > a');
```
drishu: isn't it sufficient to call li.active to fetch the active link?

modules/oe_showcase_search/tests/src/ExistingSite/IntegrationTest.php
```diff
@@ -225,32 +225,54 @@ protected function assertSearchResults(array $expected): void {
...
+      $last_link = NULL;
```
drishu: ->find returns null if not found no ?
combined with above comment I think we can skip the if else and just grap the 2 things, then check for null or something

modules/oe_showcase_search/tests/src/ExistingSite/IntegrationTest.php
```diff
@@ -225,32 +225,54 @@ protected function assertSearchResults(array $expected): void {
...
    */
-  protected function assertPagerLinks(array $expected, int $active_item_index = 0): void {
-    if ($expected === []) {
-      $this->assertSession()
-        ->elementNotExists('css', 'ul.pagination');
+  protected function assertPager(int $total, int $active = 0): void {
+    $pagination = $this->assertSession()
+      ->elementExists('css', 'ul.pagination');
+
+    $get_page_index = static function (NodeElement $link): string {
```
drishu: I agree that I would have also put it in a helper function, named or not, both are fine for me here, but do try to stick to certain scenarios when using them:
1. when there is no reusability within your project/you use it only in that specific place
2. as value of a callable parameter

at least this is my compass for it

drishu (changes requested):

modules/oe_showcase_search/tests/src/ExistingSite/IntegrationTest.php
donquixote: in this case the active and last link are identical. i want to verify that this is the case.

drishu: you check the page index of the active link and then separately  for the last link, I dont see why you would care if its the same, but I wont insist, end result is the same so up 2 u

modules/oe_showcase_search/tests/src/ExistingSite/IntegrationTest.php
```diff
@@ -225,32 +225,65 @@ protected function assertSearchResults(array $expected): void {
...
+
+  /**
+   * Asserts the page index in a pager link.
+   *
+   * @param int $expected
+   *   Expected page number.
+   * @param \Behat\Mink\Element\NodeElement $pager_link
+   *   A link element that belongs to a pager.
+   */
+  protected function assertPagerLinkPageNumber(int $expected, NodeElement $pager_link): void {
```
drishu, line 265: I would have expected a string and typecasted when sending the value but not a biggy

drishu (dismissed):

modules/oe_showcase_search/tests/src/ExistingSite/IntegrationTest.php
donquixote: I'd say semantically it is a number, and the `int` is more restrictive so gives a better idea what kind of value we expect.

modules/oe_showcase_search/tests/src/ExistingSite/IntegrationTest.php
donquixote: I agree now.
My earlier motivation was that I don't fully trust the pager, and wanted to be sure the last item is really the last link with the icon, not a bug where the last link is missing for some reason. This is why I wanted to check labels _and_ link destination.
But we cannot reliably check this anyway. So now I drop the label checks.

modules/oe_showcase_search/tests/src/ExistingSite/IntegrationTest.php
donquixote: Also I wanted to get the typecast out of the calling code, e.g. no `(string) ($total - 1)`.

drishu (dismissed):

escuriola (dismissed):

brummbar (dismissed):

donquixote:
Pushed again, removed the commit in composer.json

drishu (approved):

brummbar (approved):

PR #135, opened 2022-05-19
OEL-1614: Permissions and roles for pages and contact forms

donquixote:


tests/src/Traits/PathAccessTrait.php
```diff
@@ -0,0 +1,85 @@
+<?php
+
+declare(strict_types=1);
```
drishu: how did this pass phpcs ?

config/install/user.role.manage_contact_forms.yml
```diff
@@ -0,0 +1,15 @@
...
+  module:
+    - contact
+    - contact_storage
+    - contact_storage_export
+id: manage_contact_forms
+label: 'Manage contact forms'
+weight: 4
+is_admin: null
+permissions:
+  - 'administer contact forms'
```
drishu, line 18: the user has to know the urls by heart, lets add some permissions to view the admin pages, access toolbar

config/install/user.role.manage_contact_forms.yml
```diff
@@ -0,0 +1,15 @@
...
+  - 'export contact form messages'
+  - 'view contact messages'
```
drishu, line 20: we also agreed that this role (and only this role) would have the "access site-wide contact form", this way after creating the form he/she can already view the form and embed once good and ready

drishu (changes requested):
round 1

drishu (dismissed):

tests/src/Traits/PathAccessTrait.php
```diff
@@ -0,0 +1,84 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\Tests\oe_showcase\Traits;
+
+/**
+ * Provides methods to check access to paths.
+ */
+trait PathAccessTrait {
```
brummbar: This trait contains path assertions, let's rename it accordingly:

```suggestion
trait AssertPathAccessTrait {
```

tests/src/Traits/PathAccessTrait.php
```diff
@@ -0,0 +1,84 @@
...
+
+  /**
+   * Asserts that only users with given role can visit given paths.
+   *
+   * Side effects:
+   * - Creates user accounts.
+   * - Changes login state.
+   * - Changes currently visited page.
```
brummbar: This can be seen in the code easily. Let's drop all of these entries, also in the other methods.

tests/src/Traits/PathAccessTrait.php
```diff
@@ -0,0 +1,84 @@
...
+    $this->drupalLogout();
+    $this->assertPathsResponseCode(403, $paths);
+
+    // Test users with other roles.
+    $all_roles = array_keys(user_role_names(TRUE));
+    $other_roles = array_diff($all_roles, [$role, 'authenticated']);
+    $this->drupalLogin($this->createUserWithRoles($other_roles));
+    $this->assertPathsResponseCode(403, $paths);
+
+    // Test editor.
```
brummbar:
```suggestion
    // Test the specified role.
```

brummbar: Just as a note, it might be that in the future multiple roles can access a route, so we will have to change it.
Since we don't have such scenario now, let's leave this like it is now.

brummbar (changes requested):
I like the tests!

tests/src/Traits/PathAccessTrait.php
donquixote: A method name + doc comment should reflect how the method behaves to the outside.
If the name starts with `assert`, I would typically expect a pure method with no side effect.
This is not the case here, the behavior defies expectation for an assert method - this is worthwhile to mention in a  comment.
If you can propose a different name, I am ready to drop the comment :)

tests/src/Traits/PathAccessTrait.php
donquixote: +1 for the comment fix.

> Just as a note, it might be that in the future multiple roles can access a route, so we will have to change it.
> Since we don't have such scenario now, let's leave this like it is now.

I was thinking the same.
We should change it based on an actual scenario, otherwise we risk to add a useless dimension of flexibility.

tests/src/Traits/PathAccessTrait.php
brummbar: Uhm let's keep it indeed.

brummbar (approved):

drishu (approved):

PR #136, opened 2022-05-20
OEL-1620: Revoke administer nodes permission to editor role.

escuriola:
### Description

Revoke administer nodes permission to editor role

### Change log

- Added:
- Changed: Editor role
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/oe_showcase.post_update.php
+++ b/oe_showcase.post_update.php
@@ -9,6 +9,7 @@ declare(strict_types=1);
 
 use Drupal\block\Entity\Block;
 use Drupal\oe_bootstrap_theme\ConfigImporter;
+use Drupal\user\Entity\Role;
 
 /**
  * Set the updated regions of the oe_whitelabel navigation blocks.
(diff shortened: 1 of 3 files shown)


oe_showcase.post_update.php
```diff
@@ -62,8 +63,9 @@ function oe_showcase_post_update_00003(&$sandbox) {
...
 
-  $configs = [
-    'user.role.manage_users',
-  ];
-  ConfigImporter::importMultiple('profile', 'oe_showcase', '/config/post_updates/00003_manage_users', $configs);
+  ConfigImporter::importSingle('profile', 'oe_showcase', '/config/post_updates/00003_manage_users', 'user.role.manage_users');
+
+  $role = Role::load('editor');
+  $role->revokePermission('administer nodes');
+  $role->save();
```
donquixote, line 78: Let's make a separate update hook. This also gives you more space for a description.
Background: We recently decided we want separate update hooks rather than reusing the same.

tests/src/ExistingSite/ShowcaseExistingSiteCreateNewsTest.php
```diff
@@ -76,7 +76,6 @@ public function testCreateNews() {
     $page->fillField('Content', 'Example Content');
     $page->fillField('Introduction', 'Example Introduction');
     $page->fillField('Date', '2022-01-24');
-    $page->fillField('Time', '00:00:00');
```
donquixote, line 79: This is the one I found as well. I thought there might be more, but if our tests are green then this is good enough :)
We can add more tests in follow-up issues.

(resolve after reading)

oe_showcase.post_update.php
donquixote: If you do, let's keep the other method untouched. It is already approved and we avoid unnecessary risk by not touching other upgrade hooks.

drishu (approved):

donquixote (approved):

PR #137, opened 2022-05-20
OEL-1618: Upgrade to BCL 0.23.0.

escuriola:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -51,7 +51,7 @@
         "openeuropa/oe_paragraphs": "^1.13",
         "openeuropa/oe_starter_content": "1.0.0-beta1",
         "openeuropa/oe_webtools": "^1.18",
-        "openeuropa/oe_whitelabel": "1.0.0-beta3"
+        "openeuropa/oe_whitelabel": "0.1618.202206241439"
     },
     "require-dev": {
         "composer/installers": "~1.11",
--- a/modules/oe_showcase_page/templates/oe-webtools-social-share.html.twig
+++ b/modules/oe_showcase_page/templates/oe-webtools-social-share.html.twig
@@ -9,7 +9,7 @@
 <div class="bcl-social-media-share">
   {% include '@oe-bcl/bcl-heading/heading.html.twig' with {
     title: title,
-    title_tag: 'h3',
+    title_tag: 'h2',
   } only %}
   <script type="application/json">{{ icons_json }}</script>
 </div>
--- a/modules/oe_showcase_page/templates/oe-webtools-social-share.html.twig
+++ b/modules/oe_showcase_page/templates/oe-webtools-social-share.html.twig
@@ -7,6 +7,9 @@
  */
 #}
 <div class="bcl-social-media-share">
-  <h3 class="fw-bold">{{ title }}</h3>
+  {% include '@oe-bcl/bcl-heading/heading.html.twig' with {
+    title: title,
+    title_tag: 'h3',
+  } only %}
   <script type="application/json">{{ icons_json }}</script>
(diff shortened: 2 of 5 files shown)


donquixote:
We must upgrade composer.lock or it won't work.

donquixote:
> We must upgrade composer.lock or it won't work.
```
docker-compose exec web composer update openeuropa/oe_whitelabel openeuropa/oe_bootstrap_theme
```
This also explains the failing test.

donquixote:
I notice we have some templates where we put custom hx elements with fw-bold class.
I found these by looking for "fw-bold" in the few templates we have in showcase.

```
modules/oe_showcase_page/templates/oe-webtools-social-share.html.twig
modules/oe_showcase_user_profile/templates/user--full.html.twig
```

In netlify these are not bold, from what I found.

TBD if we do this in a follow-up or here.

donquixote:
I still see fw-bold here.

modules/oe_showcase_user_profile/templates/user--full.html.twig
```diff
@@ -35,7 +35,10 @@
       }]) %}
     {% endfor %}
     {% if _items %}
-      <h3 class="fw-bold mb-4">{{ group_label }}</h3>
+      {% include '@oe-bcl/bcl-heading/heading.html.twig' with {
+        title: group_label,
+        title_tag: 'h3',
+      } only %}
```
donquixote, line 41: I suspect that h2 would be more consistent here. But netlify still has h3 on user profile page, so let's keep it like that.

modules/oe_showcase_page/templates/oe-webtools-social-share.html.twig
```diff
@@ -7,6 +7,9 @@
  */
 #}
 <div class="bcl-social-media-share">
-  <h3 class="fw-bold">{{ title }}</h3>
+  {% include '@oe-bcl/bcl-heading/heading.html.twig' with {
+    title: title,
+    title_tag: 'h3',
+  } only %}
```
donquixote: In netlify I find this, https://oelibrary.netlify.app/default/?path=/story/paragraphs-social-media-follow--default but it looks a bit different because it does not use webtools. Not sure if this is the right place to look.
In netlify this has h2, not h3.

donquixote (dismissed):

composer.json
```diff
@@ -103,9 +103,6 @@
         "patches": {
             "drupal/facets": {
                 "https://www.drupal.org/project/facets/issues/3262863": "https://git.drupalcode.org/project/facets/-/commit/88fd8a88a7206ea2e3828718fc404665ee20b064.diff"
-            },
-            "openeuropa/oe_list_pages": {
-                "Support Address field in country code processor": "https://patch-diff.githubusercontent.com/raw/openeuropa/oe_list_pages/pull/139.diff"
```
brummbar: How come we are dropping this?

modules/oe_showcase_page/templates/oe-webtools-social-share.html.twig
brummbar: Actually this was asked to be changed by UX and since we are here, let's replace it with a `<div class="fs-4 fw-semibold mb-3">`.

brummbar (changes requested):

drishu (approved):

brummbar (approved):

PR #139, opened 2022-05-20
OEL-1626: Improve editor permissions

brummbar:
Add permissions to manage supported media bundles and clean some existing permissions.


tests/src/ExistingSite/MediaAccessTest.php
```diff
@@ -0,0 +1,137 @@
...
+use weitzman\DrupalTestTraits\Entity\MediaCreationTrait;
+
+/**
+ * Tests access to media CRUD and overview routes.
+ */
+class MediaAccessTest extends ShowcaseExistingSiteTestBase {
+
+  use MediaCreationTrait;
+  use PathAccessTrait;
+  use UserTrait;
```
drishu, line 19: looks a bit weird to call this here since no explicit call of the UserTrait is made here, I'd use UserTrait in PathAccessTrait and remove it from here

tests/src/ExistingSite/MediaAccessTest.php
```diff
@@ -0,0 +1,137 @@
...
+    $this->assertPathsRequireRole($restricted_paths, 'editor');
+
+    $forbidden_paths = [];
+    foreach ($disallowed_media_bundles as $bundle) {
+      $forbidden_paths[] = 'media/add/' . $bundle;
+      $forbidden_paths[] = 'media/' . $this->medias[$bundle]->id() . '/edit';
+      $forbidden_paths[] = 'media/' . $this->medias[$bundle]->id() . '/delete';
+    }
+
+    $this->assertPathsResponseCode(403, $forbidden_paths);
```
donquixote: We should test this for specific roles (user with all roles, and anon).
Currently it is not even clear which role is currently logged in.

I suggest a new method in the trait, `assertPathsAdminOnly()` with:
- anonymous access denied
- user with _all_ roles access denied
- user 1 access granted

The latter is to make sure that site builders can do their job, _and_ that the paths actually exist and are not denied for some other silly reason.

.drone.yml
```diff
@@ -71,6 +71,8 @@ pipeline:
     group: test
     image: registry.fpfis.eu/fpfis/httpd-php:${PHP_VERSION}-ci
     commands:
+      - ./vendor/bin/run ci:site-setup
```
drishu, line 74: this is nice

drishu (changes requested):

tests/src/ExistingSite/MediaAccessTest.php
```diff
@@ -0,0 +1,137 @@
...
+        ],
+      ],
+    ];
+
+    foreach ($bundles as $bundle => $data) {
+      $this->medias[$bundle] = $this->createMedia($data + [
+        'bundle' => $bundle,
+        'name' => $bundle . ' title',
+      ]);
+    }
```
donquixote, line 100: Should this be in setUp()? Or can it be in a helper method that returns the media entities?

donquixote: @drishu when reviewing can you test if the medias are cleaned up after the test has run?

tests/src/Traits/AssertPathAccessTrait.php
```diff
@@ -59,6 +59,28 @@ protected function assertPathsRequireRole(array $paths, string $role): void {
...
+   *
+   * Side effects:
+   * - Creates user accounts.
+   * - Changes login state.
+   * - Changes currently visited page.
+   *
+   * @param string[] $paths
+   *   The paths to test.
+   */
+  protected function assertPathsAccessForbidden(array $paths): void {
```
donquixote, line 73: I would name it assertPathsAccessDenied(), but it's ok, no reason to block it.

tests/src/ExistingSite/MediaAccessTest.php
brummbar: No we shoudln't do that, using traits inside traits becomes bad very quickly.
Imagine a second trait doing the same. You cannot include them together anymore.

tests/src/ExistingSite/MediaAccessTest.php
brummbar: This is in the setup indeed.

tests/src/ExistingSite/MediaAccessTest.php
brummbar: Aside from uid 1, I agree.

tests/src/ExistingSite/MediaAccessTest.php
brummbar: Also using the createMedia() already marks all the medias.

tests/src/Traits/AssertPathAccessTrait.php
```diff
@@ -59,6 +59,28 @@ protected function assertPathsRequireRole(array $paths, string $role): void {
...
+    // Test anonymous.
+    $this->drupalLogout();
+    $this->assertPathsResponseCode(403, $paths);
+
+    // Test all existing roles.
+    $all_roles = array_keys(user_role_names(TRUE));
```
donquixote: We need to remove 'authenticated'.

tests/src/Traits/AssertPathAccessTrait.php
brummbar: "403 Forbidden" is the status code.

tests/src/ExistingSite/MediaAccessTest.php
```diff
@@ -0,0 +1,137 @@
...
+
+use Drupal\file\Entity\File;
+use Drupal\Tests\oe_showcase\Traits\AssertPathAccessTrait;
+use Drupal\Tests\oe_showcase\Traits\UserTrait;
+use weitzman\DrupalTestTraits\Entity\MediaCreationTrait;
+
+/**
+ * Tests access to media CRUD and overview routes.
+ */
+class MediaAccessTest extends ShowcaseExistingSiteTestBase {
```
donquixote, line 15: A bit weird that we have a separate test class for media, but node paths are part of AuthorisationTest.
But ok to keep it like this and rethink later.

brummbar: I prefer code separation for the creation of medias and such.

tests/src/ExistingSite/MediaAccessTest.php
drishu: "Imagine a second trait doing the same. You cannot include them together anymore." this bit is not true

donquixote: Trait in trait is used many times in core
```
ag "  use .*Trait;" build/core/**/*Trait.php
```
and works fine, see https://3v4l.org/pDCCO

donquixote: Conflicts occur if 2 traits provide a method with the same name. But not if they inherit this method from the same other trait.

tests/src/ExistingSite/MediaAccessTest.php
donquixote: > This is in the setup indeed.

is/ought

drishu (approved):

tests/src/Traits/AssertPathAccessTrait.php
donquixote: yes. but for me it would either be assertPathsForbidden() or assertPathsAccessDenied(), not access + forbidden.
But let's keep it like it is for now.

donquixote (approved):

tests/src/ExistingSite/MediaAccessTest.php
brummbar: Both, as we discussed already multiple times this is how I see it: we have a test class that, in the current status, needs the medias created for each of its test methods (which is one).
So it can be in the setup.

PR #142, opened 2022-05-24
OEL-1169: OE Webtools ETrans integration

escuriola:
### Description

Enable oe_webtools_etrans module and block on the footer region

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


modules/oe_showcase_multilingual/config/optional/block.block.openeuropa_webtools_etrans.yml
```diff
@@ -0,0 +1,36 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - node
+    - oe_webtools_etrans
+  theme:
+    - oe_whitelabel
+id: openeuropa_webtools_etrans
```
drishu: lets make it showcase_etrans

modules/oe_showcase_multilingual/config/post_updates/00001_webtools_etrans/block.block.openeuropa_webtools_etrans.yml
```diff
@@ -0,0 +1,36 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - node
+    - oe_webtools_etrans
+  theme:
+    - oe_whitelabel
+id: openeuropa_webtools_etrans
```
drishu: showcase_etrans

modules/oe_showcase_multilingual/tests/src/ExistingSiteJavascript/WebToolsEtransTest.php
```diff
@@ -0,0 +1,71 @@
...
+    $this->assertStringContainsString(
+      '{"service":"etrans","languages":{"exclude":["' . $current_language . '"]},"renderAs":{"button":false,"icon":false,"link":true},"domain":"spd","delay":0,"renderTo":"webtools-etrans","include":"main, .bcl-header"}',
+      $etrans_block->getHtml()
+    );
+
+    // Assert excluded language after changing the current language.
+    $current_language = 'es';
+    $this->clickLink('English');
+    $this->assertSession()->waitForElementVisible('css', '#languageModal');
+    $this->clickLink('español');
```
drishu: just visit `$this->drupalGet('/es/node/' . $node->id());`

modules/oe_showcase_multilingual/tests/src/ExistingSiteJavascript/WebToolsEtransTest.php
```diff
@@ -0,0 +1,71 @@
...
+    $etrans_block = $page->find('css', 'div.page > div#block-openeuropa-webtools-etrans');
+    $assert_session->waitForId('wtEtransRequester');
+    $etrans_link = $etrans_block->find('css', 'div.etrans.wt.wt-etrans  > div#wtEtransRequester > a.wt-link');
+    $this->assertEquals('Translate this page', $etrans_link->getText());
+    $this->assertStringContainsString(
+      '{"service":"etrans","languages":{"exclude":["' . $current_language . '"]},"renderAs":{"button":false,"icon":false,"link":true},"domain":"spd","delay":0,"renderTo":"webtools-etrans","include":"main, .bcl-header"}',
+      $etrans_block->getHtml()
+    );
+
+    // Assert excluded language after changing the current language.
```
drishu: what do you mean by excluded language?

modules/oe_showcase_multilingual/tests/src/ExistingSiteJavascript/WebToolsEtransTest.php
```diff
@@ -0,0 +1,71 @@
...
+    $etrans_block = $page->find('css', 'div.page > div#block-openeuropa-webtools-etrans');
+    $assert_session->waitForId('wtEtransRequester');
+    $etrans_link = $etrans_block->find('css', 'div.etrans.wt.wt-etrans  > div#wtEtransRequester > a.wt-link');
+    $this->assertEquals('Traducir esta página', $etrans_link->getText());
+    $this->assertStringContainsString(
+      '{"service":"etrans","languages":{"exclude":["' . $current_language . '"]},"renderAs":{"button":false,"icon":false,"link":true},"domain":"spd","delay":0,"renderTo":"webtools-etrans","include":"main, .bcl-header"}',
+      $etrans_block->getHtml()
+    );
+
+    // Assert WebToolsEtrans Block is not visible out of a node.
```
drishu:
```suggestion
    // Assert Etrans is not visible outside of nodes.
```
Something like this

modules/oe_showcase_multilingual/tests/src/ExistingSiteJavascript/WebToolsEtransTest.php
```diff
@@ -0,0 +1,71 @@
...
+    $page = $this->getSession()->getPage();
+
+    $user = $this->createUser([
+      'create oe_showcase_page content',
+      'delete own oe_showcase_page content',
+      'edit own oe_showcase_page content',
+    ]);
+    $this->drupalLogin($user);
+
+    // Assert WebToolsEtrans Block exists if node and is placed in the footer.
```
drishu:
```suggestion
    // Assert Etrans is loaded for nodes.
```

modules/oe_showcase_multilingual/tests/src/ExistingSiteJavascript/WebToolsEtransTest.php
```diff
@@ -0,0 +1,71 @@
...
+    $this->assertEquals('Traducir esta página', $etrans_link->getText());
+    $this->assertStringContainsString(
+      '{"service":"etrans","languages":{"exclude":["' . $current_language . '"]},"renderAs":{"button":false,"icon":false,"link":true},"domain":"spd","delay":0,"renderTo":"webtools-etrans","include":"main, .bcl-header"}',
+      $etrans_block->getHtml()
+    );
+
+    // Assert WebToolsEtrans Block is not visible out of a node.
+    $this->drupalGet('/user/' . $user->id());
+    $assert_session->elementNotExists('css', 'div#block-openeuropa-webtools-etrans');
+  }
```
drishu: Not sure about the test, we should revisit after these major changes

drishu (changes requested):
- the etrans should appear in the language switcher next to the dropdown
- we should override the template like here https://github.com/ec-europa/ewcms/blob/c054179566b25e6b80c4f063db83394e82c9d83e/themes/ewcms_theme/templates/monolingual-etrans-block.html.twig
- ideally it should appear only when a translation is not available

drishu (dismissed):

composer.json
```diff
@@ -92,6 +92,11 @@
         "composer-exit-on-patch-failure": true,
         "enable-patching": true,
         "installer-types": ["npm-asset", "bower-asset"],
+        "patches": {
+            "openeuropa/oe_webtools": {
+                "oe_webtools_etrans": "https://patch-diff.githubusercontent.com/raw/openeuropa/oe_webtools/pull/178.patch"
```
brummbar: This has been merged into master, let's use that branch.

modules/oe_showcase_multilingual/oe_showcase_multilingual.module
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Implements hook_block_access().
+ */
+function oe_showcase_multilingual_block_access(Block $block, $operation, AccountInterface $account) {
+  if ($operation !== 'view' || $block->id() !== 'showcase_etrans') {
+    return AccessResult::neutral();
+  }
+
+  /** @var \Drupal\node\NodeInterface $node */
+  $node = \Drupal::routeMatch()->getParameter('node');
```
brummbar, line 30: This is quite loose, the best would be to use the node context, but it's ok for now.

modules/oe_showcase_multilingual/oe_showcase_multilingual.module
```diff
@@ -0,0 +1,34 @@
...
+  if (!$node) {
+    return AccessResult::neutral();
```
brummbar: We are missing the cache context 'route'.

modules/oe_showcase_multilingual/oe_showcase_multilingual.module
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  $translations = $node->getTranslationLanguages(FALSE);
+  if (empty($translations)) {
+    return AccessResult::allowed();
+  }
+
+  return AccessResult::forbidden();
```
brummbar: A few things:

1. We should add the node as cache dependency. Everything works now, because the context adds it, but better to be explicit.
2. We should return forbidden if the translations are empty. Otherwise we should return neutral, and leave the applied contexts to determine the result. By returning allowed, we are saying "show it even if the context don't match". Again it's ok in this case, but better not to.

modules/oe_showcase_multilingual/tests/src/ExistingSite/WebToolsEtransTest.php
```diff
@@ -0,0 +1,61 @@
...
+      'type' => 'oe_showcase_page',
+      'title' => 'Test page',
+      'body' => 'This profile is aimed to provide developers with a series of examples on how the OpenEuropa Library features are built and configured.',
+      'status' => 1,
+      'language' => 'en',
+    ];
+    $node = $this->createNode($values);
+    $this->drupalGet($node->toUrl());
+
+    $etrans_block = $page->find('css', 'main > div.container > div.row > div.col-12 > div#block-showcase-etrans');
```
brummbar:
```suggestion
    $etrans_block = $assert_session->elementExists('css', 'main > div.container > div.row > div.col-12 > div#block-showcase-etrans');
```
So we can drop the $page too.

modules/oe_showcase_multilingual/tests/src/ExistingSite/WebToolsEtransTest.php
```diff
@@ -0,0 +1,61 @@
...
+    $this->assertStringContainsString(
+      '{"service":"etrans","languages":{"exclude":["en"]},"renderAs":{"button":false,"icon":false,"link":true},"domain":"spd","delay":0,"renderTo":"webtools-etrans","include":"main, .bcl-header"}',
+      $etrans_block->getHtml()
+    );
+
+    // Assert Etrans is not loaded for nodes with translations.
+    $node_es = $node->addTranslation('es', $node->toArray());
+    $node_es->setTitle('Spanish Translation');
+    $node_es->save();
+
```
brummbar: Unnecessary space.

modules/oe_showcase_multilingual/tests/src/ExistingSite/WebToolsEtransTest.php
```diff
@@ -0,0 +1,61 @@
...
+    $this->drupalGet($node->toUrl());
+
```
brummbar: Unnecessary space.

modules/oe_showcase_multilingual/tests/src/ExistingSite/WebToolsEtransTest.php
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * Tests oe_webtools_etrans block is displaying properly.
+ */
+class WebToolsEtransTest extends ShowcaseExistingSiteTestBase {
+
+  /**
+   * Test oe_webtools_etrans block in oe_showcase.
+   */
+  public function testWebToolsEtrans(): void {
+    $this->markEntityTypeForCleanup('node');
```
brummbar: Not needed. createNode() marks them already.

brummbar (changes requested):
For me the link was unstyled, is it correct?
Maybe we should ask BAs and UX to place this slightly better.

modules/oe_showcase_multilingual/oe_showcase_multilingual.module
donquixote: It will/would also match on node/*/edit pages or other node routes.
But I think this is excluded in the block visibility settings, and anyway these pages use the back-end theme.
So yeah seems ok for now.

modules/oe_showcase_multilingual/oe_showcase_multilingual.module
donquixote: > We should return forbidden if the translations are empty.
I think you mean the opposite.

modules/oe_showcase_multilingual/config/post_updates/00001_webtools_etrans/block.block.showcase_etrans.yml
```diff
@@ -0,0 +1,36 @@
...
+status: true
+dependencies:
+  module:
+    - node
+    - oe_webtools_etrans
+  theme:
+    - oe_whitelabel
+id: showcase_etrans
+theme: oe_whitelabel
+region: content
```
donquixote, line 11: Issue said to put it in the footer.
But I assume this placement has been agreed on. If so then ok for me.

modules/oe_showcase_multilingual/config/post_updates/00001_webtools_etrans/block.block.showcase_etrans.yml
```diff
@@ -0,0 +1,36 @@
...
+plugin: oe_webtools_etrans
+settings:
+  id: oe_webtools_etrans
+  label: 'OpenEuropa Webtools eTrans'
+  label_display: '0'
+  provider: oe_webtools_etrans
+  delay: 0
+  domain: spd
+  exclude: ''
+  include: 'main, .bcl-header'
```
donquixote, line 23: I know this was requested in the issue summary.
To me it seems a bit strange that we make the translation dependent on the node translate status, but then we also translate the header.
But ok for me considering it has been decided like this.

donquixote (dismissed):
Seems alright to me, but needs rebase and resolving of conflicts.
I am not going to merge it myself, I was not enough involved in the issue.

composer.json
```diff
@@ -47,7 +47,7 @@
         "openeuropa/oe_multilingual": "^1.9",
         "openeuropa/oe_paragraphs": "^1.13",
         "openeuropa/oe_starter_content": "1.0.0-beta1",
-        "openeuropa/oe_webtools": "^1.14",
+        "openeuropa/oe_webtools": "^1.17",
```
drishu: you changed the webtool version but this is not reflected in the lock;
also bump up to 1.18, the patch does not install anymore.

drishu (changes requested):

drishu (approved):

brummbar (approved):

PR #143, opened 2022-05-24
OEL-1571: Improve Contact Forms testing.

escuriola:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/modules/oe_showcase_contact_forms/tests/src/ExistingSite/ContactFormTest.php
+++ b/modules/oe_showcase_contact_forms/tests/src/ExistingSite/ContactFormTest.php
@@ -91,7 +91,6 @@ class ContactFormTest extends ShowcaseExistingSiteTestBase {
     $page->fillField('Phone', '345345345');
     $page->selectFieldOption('Country of residence', 'http://publications.europa.eu/resource/authority/country/BEL');
     $page->pressButton('Send message');
-    $assert_session->pageTextContains('Belgium');
 
     $confirm_message = $page->find('css', 'div.alert.alert-success');
     $this->assertStringContainsString('Alpaca', $confirm_message->getText());
--- a/modules/oe_showcase_contact_forms/tests/src/ExistingSite/ContactFormTest.php
+++ b/modules/oe_showcase_contact_forms/tests/src/ExistingSite/ContactFormTest.php
@@ -48,7 +48,7 @@ class ContactFormTest extends ShowcaseExistingSiteTestBase {
     $this->markEntityTypeForCleanup('node');
     $this->markEntityTypeForCleanup('paragraph');
 
-    // Disable the flood limits for contact forms.
+    // Disable the flood limits to avoid submit failure when run multiple times.
     $contact_settings = $this->container->get('config.factory')->getEditable('contact.settings');
     $flood_values = [
       'limit' => '100',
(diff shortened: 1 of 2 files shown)


modules/oe_showcase_contact_forms/tests/src/ExistingSite/ContactFormTest.php
```diff
@@ -2,7 +2,7 @@
 
 declare(strict_types = 1);
 
-namespace Drupal\Tests\oe_showcase_contact_forms\ExistingSite;
+namespace Drupal\Tests\oe_showcase_contact_forms;
```
donquixote: Why?
We are in `/ExistingSite/` subdir, so this should also be reflected in the namespace.
Looks like the namespace was already fixed here, a5c64c6db430aa810abe4ebba84844836c4daed0

modules/oe_showcase_contact_forms/tests/src/ExistingSite/ContactFormTest.php
```diff
@@ -74,10 +92,12 @@ public function testCreateContactForm() {
...
-    $assert_session->pageTextContains('Alpaca');
-    $assert_session->pageTextContains('345345345');
-    $assert_session->pageTextContains('Belgium');
-    $assert_session->pageTextContains('Example Message text');
+    $confirm_message = $page->find('css', 'div.alert.alert-success');
+    $this->assertStringContainsString('Alpaca', $confirm_message->getText());
+    $this->assertStringContainsString('345345345', $confirm_message->getText());
+    $this->assertStringContainsString('Belgium', $confirm_message->getText());
+    $this->assertStringContainsString('Example subject', $confirm_message->getText());
+    $this->assertStringContainsString('Example Message text', $confirm_message->getText());
```
donquixote: If we fully rely on ->getText(), we could call it once and have a local var that only contains the text, instead of a NodeElement.
```php
    $confirm_message = $page->find('css', 'div.alert.alert-success')->getText();
```

modules/oe_showcase_contact_forms/tests/src/ExistingSite/ContactFormTest.php
donquixote: I notice the Jira issue says we should move it to `Drupal\\Tests\\oe_showcase_contact_forms`. This was wrong advice :)

donquixote (dismissed):

donquixote (dismissed):

modules/oe_showcase_contact_forms/tests/src/ExistingSite/ContactFormTest.php
```diff
@@ -38,6 +48,14 @@ public function testCreateContactForm() {
     $this->markEntityTypeForCleanup('node');
     $this->markEntityTypeForCleanup('paragraph');
 
+    // Disable the flood limits for contact forms.
```
drishu: There are two good reasons to add a comment:
1) the code is unclear, obfuscated and need clarifications;
here we add a comment explaining what the code does.
2) the code is clear but the reason for adding it is not;
here we add a comment why we added the code, the reason/motivation.

Here we are in the second situation, what the code does is clear, but the reason for adding it is not:
```suggestion
    // Disable the flood limits to avoid submit failure when run multiple times.
```

drishu (changes requested):

drishu (dismissed):

drishu (dismissed):

donquixote (approved):

drishu (approved):

PR #144, opened 2022-05-25
OEL-1418: Update oe_showcase after remove paragraphs required components on oe_bootstrap_theme.

escuriola:
…

### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,7 +44,7 @@
         "openeuropa/oe_paragraphs": "^1.13",
         "openeuropa/oe_starter_content": "1.0.0-beta1",
         "openeuropa/oe_webtools": "^1.14",
-        "openeuropa/oe_whitelabel": "1.0.0-beta2"
+        "openeuropa/oe_whitelabel": "0.1418.202205311230"
     },
     "require-dev": {
         "composer/installers": "~1.11",


composer.json
```diff
@@ -39,7 +39,7 @@
         "openeuropa/oe_content": "^2.8.0",
         "openeuropa/oe_corporate_blocks": "^4.4",
         "openeuropa/oe_corporate_countries": "^2.0.0-alpha3",
-        "openeuropa/oe_media": "^1.14",
+        "openeuropa/oe_media": "^1.16",
```
drishu: what for?

escuriola: Because is the oe_bootstrap_theme 1.x required version: https://github.com/openeuropa/oe_bootstrap_theme/blob/1.x/composer.json

So as I removed there, I updated in the other components.

drishu: no, see explanation in whitelabel PR

composer.json
```diff
@@ -106,6 +106,14 @@
                 }
             }
         },
+        "patches": {
+            "openeuropa/oe_bootstrap_theme": {
+                "latest": "https://github.com/openeuropa/oe_bootstrap_theme/compare/1.0.0-beta1..OEL-1418.diff"
```
drishu: we cant patch, we have composer changes, use a whitelabe dev tag

drishu (approved):

escuriola:
PR only for testing purposes

PR #148, opened 2022-05-31
OEl-1652: Test Drupal upgrade paths

drishu:
### Description

Automate upgrade testing by leveraging 

### Change log

- Added: drone steps
- Changed: drone matrix

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -58,7 +58,7 @@ pipeline:
       - /cache:/cache
     commands:
       - git config --global --add safe.directory /test/oe_showcase
-      - currentbranch=$(git branch --show-current)
+      - git fetch --tags
       - git checkout $(git describe --tags $(git rev-list --tags --max-count=1))
       - composer install --ansi --no-progress
     when:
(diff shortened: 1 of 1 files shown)


.drone.yml
```diff
@@ -79,5 +122,6 @@ pipeline:
         exclude: tag
```
drishu: not in scope, but was thinking, shouldn't we leave the install test to run when building a tag (so remove the exludes from install-branch, grumphp, and phpunit)?

.drone.yml
```diff
@@ -41,26 +41,69 @@ services:
...
+      - git fetch --all --tags
+      - git checkout $(git describe --tags $(git rev-list --tags --max-count=1))
+      - composer install --ansi --no-progress
+    when:
+      matrix:
+        TYPE: update
+      event:
+        exclude: tag
+
+  install-branch:
```
drishu, line 68: should we invest in a task that sets up the needed config in settings.php so we dont run the full site-install on the branch when in the update scenario ?

escuriola (dismissed):

.drone.yml
```diff
@@ -79,5 +134,6 @@ pipeline:
         exclude: tag
 
 matrix:
-  PHP_VERSION:
-    - 8.0
+  include:
```
brummbar: We don't need the include.

brummbar (changes requested):

brummbar (approved):

escuriola (approved):

PR #149, opened 2022-06-01
OEL-1651: Get the oe_starter_content development branch to test the published date

escuriola:
### Description

Get the oe_starter_content development branch to test the published date. If published date is empty, then add the current date as published date.

### Change log

- Added:
- Changed: oe_starter_content has been updated
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -106,6 +106,9 @@
             },
             "openeuropa/oe_list_pages": {
                 "Support Address field in country code processor": "https://patch-diff.githubusercontent.com/raw/openeuropa/oe_list_pages/pull/139.diff"
+            },
+            "openeuropa/oe_starter_content": {
+                "latest": "https://github.com/openeuropa/oe_starter_content/compare/1.0.0-beta1...1.x.diff"
             }
         },
         "artifacts": {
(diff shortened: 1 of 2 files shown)


donquixote:
I talked with Francesco.
We should also test the empty date in ShowcaseExistingSiteCreateNewsTest, because in showcase we have (or will have) a mechanism in place to test the upgrade path. So this is a convenient way to confirm that the field default value setting will be updated via the update hook in oe_starter_content.

I suggest:
- visit `node/add/oe_sc_news`, as we already do
- verify that the field is pre-filled (new). not sure how to check that. no need to check the exact date.
- set the field to a custom value, as we already do

this seems the easiest way to do it, and we don't have to go back to the edit page to save it again.
but if you have a better idea let me know :)

donquixote (dismissed):

drishu (dismissed):

brummbar (approved):
Re-approving on behalf of @drishu .

donquixote (approved):

PR #153, opened 2022-06-07
OEL-1675: News title facet should search in all fields.

drishu:
### Description

Requirements ask for it to be search in title, synopsis, and content.

The other way around for event, now that we add new fulltext fields, we need to make sure that it does not break the Event name search by searching in these new fields.

### Change log

- Added: fields body and oe_summary to the index
- Changed:
  - news title facet - renamed label
  - event title facet - do not search in all fulltest fields, only title
  - listed the new fields in the lowercase processors
  - adapted the tests

### DIFF
--- a/modules/oe_showcase_list_pages/config/install/facets.facet.oelp_oe_sc_event__oe_sc_event_dates.yml
+++ b/modules/oe_showcase_list_pages/config/install/facets.facet.oelp_oe_sc_event__oe_sc_event_dates.yml
@@ -3,27 +3,27 @@ status: true
 dependencies: {  }
 id: oelp_oe_sc_event__oe_sc_event_dates
 name: 'Event dates'
+url_alias: oelp_oe_sc_event__oe_sc_event_dates
 weight: 0
 min_count: 1
-url_alias: oelp_oe_sc_event__oe_sc_event_dates
-facet_source_id: 'list_facet_source:node:oe_sc_event'
+show_only_one_result: false
 field_identifier: oe_sc_event_dates
-query_operator: or
-hard_limit: 0
-exclude: false
-use_hierarchy: false
-keep_hierarchy_parents_active: false
-expand_hierarchy: false
-enable_parent_when_child_gets_disabled: true
+facet_source_id: 'list_facet_source:node:oe_sc_event'
 widget:
   type: oe_list_pages_date
   config:
     date_type: date
-empty_behavior:
-  behavior: none
+query_operator: or
+use_hierarchy: false
+keep_hierarchy_parents_active: false
+hierarchy:
+  type: taxonomy
+  config: {  }
+expand_hierarchy: false
+enable_parent_when_child_gets_disabled: true
+hard_limit: 0
+exclude: false
 only_visible_when_facet_source_is_visible: true
-show_only_one_result: false
-show_title: false
 processor_configs:
   active_widget_order:
     processor_id: active_widget_order
@@ -49,3 +49,6 @@ processor_configs:
       pre_query: 50
       build: 15
     settings: {  }
+empty_behavior:
+  behavior: none
+show_title: false
--- a/modules/oe_showcase_list_pages/config/install/facets.facet.oelp_oe_sc_event__title.yml
+++ b/modules/oe_showcase_list_pages/config/install/facets.facet.oelp_oe_sc_event__title.yml
@@ -13,12 +13,15 @@ hard_limit: 0
 exclude: false
 use_hierarchy: false
 keep_hierarchy_parents_active: false
+hierarchy:
+  type: taxonomy
+  config: {  }
 expand_hierarchy: false
 enable_parent_when_child_gets_disabled: true
 widget:
   type: oe_list_pages_fulltext
   config:
-    fulltext_all_fields: true
+    fulltext_all_fields: false
 empty_behavior:
   behavior: none
 only_visible_when_facet_source_is_visible: true
@@ -43,6 +46,11 @@ processor_configs:
       sort: 40
     settings:
       sort: ASC
+  hierarchy_processor:
+    processor_id: hierarchy_processor
+    weights:
+      build: 100
+    settings: {  }
   url_processor_handler:
     processor_id: url_processor_handler
     weights:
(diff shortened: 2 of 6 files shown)


modules/oe_showcase_list_pages/config/install/facets.facet.oelp_oe_sc_event__oe_sc_event_dates.yml
```diff
@@ -49,3 +49,6 @@ processor_configs:
       pre_query: 50
       build: 15
     settings: {  }
+empty_behavior:
+  behavior: none
+show_title: false
```
donquixote, line 54: Noise?

modules/oe_showcase_list_pages/config/install/facets.facet.oelp_oe_sc_event__title.yml
```diff
@@ -43,6 +46,11 @@ processor_configs:
       sort: 40
     settings:
       sort: ASC
+  hierarchy_processor:
+    processor_id: hierarchy_processor
+    weights:
+      build: 100
+    settings: {  }
```
donquixote, line 53: The config files seem to have different levels or "normalization":
After re-save + re-export (drush cde), this one has noise, the others are clean.

For this issue, we should re-save and re-export all of the modified files for a clean slate.

For future issues, I like to do this "normalization" in a separate commit, so that the actual commits are cleaner. It can be done like this:
- In a first iteration just change everything as needed and commit all at once. This is because at this time we might not know yet which configs we want to change.
- Look at the diff to identify anything that looks like "noise".
- As a second iteration, use `git rebase -ir`, put a break at the start of the branch, then import or reinstall, re-save + re-export, commit only those files that we care about, then `git rebase --continue` to replay the other changes on top of it. Resolve conflicts by prefering the new version from REBASE_HEAD.
- We end up with a branch where the first commit(s) contain the noise, and the subsequent commit(s) are reduced to meaningful intentional changes.

modules/oe_showcase_list_pages/tests/src/ExistingSite/ListPagesTest.php
```diff
@@ -223,6 +242,14 @@ public function testCreateListPages() {
...
     $search_button->click();
     $this->assertSearchResultsTitle('Event list page', 0);
+
+    // Assert Event title filters only by title.
+    $title_input->setValue('This is an Event content number 10');
+    $search_button->click();
+    $this->assertSearchResultsTitle('Event list page', 0);
+    $title_input->setValue('This is an Event introduction number 10');
+    $search_button->click();
+    $this->assertSearchResultsTitle('Event list page', 0);
```
donquixote, line 267: I think I would use less generic words (fruit, animals etc) and make at least some of them unique per node and per field. But, I think it works fine, it is good enough.

We could also test search terms in shuffled order, repeated search terms, mix of uppercase / lowercase, combination of words from different fields. E.g. "10 introduction nUmber content introduction". This would fail if we change the processors configuration for the search index or the facets, or if we add a custom processor that behaves weirdly.

I think we don't want to cover the behavior of contrib modules, but there are enough things we can mess up on our side. E.g. if we get a new version of facets or search_api, but don't properly update our settings.

But I leave it to you if we want to do that extra step (perhaps one more test is enough to cover enough of it).

modules/oe_showcase_list_pages/config/install/facets.facet.oelp_oe_sc_news__title.yml
```diff
@@ -2,25 +2,31 @@ langcode: en
 status: true
 dependencies: {  }
 id: oelp_oe_sc_news__title
-name: Title
-url_alias: oelp_oe_sc_news__title
+name: Keyword
```
donquixote: I don't like "Keyword" for this, maybe "Search terms" is better?

donquixote (changes requested):
I think we should normalize the config changes via re-save + re-export.

modules/oe_showcase_list_pages/config/install/facets.facet.oelp_oe_sc_event__oe_sc_event_dates.yml
drishu: yes

modules/oe_showcase_list_pages/config/install/facets.facet.oelp_oe_sc_news__title.yml
drishu: it used to be title, but requirements ask for keywords

modules/oe_showcase_list_pages/config/install/facets.facet.oelp_oe_sc_event__title.yml
drishu: Sounds like spending extra time on something that is not worth it, can also introduce issues when solving the conflicts.
We should rather pay attention at each config change and make sure we understand the changes.

modules/oe_showcase_list_pages/tests/src/ExistingSite/ListPagesTest.php
drishu: More randomness is a good idea, but the shuffling of words will have to do for now

modules/oe_showcase_list_pages/config/install/facets.facet.oelp_oe_sc_event__title.yml
drishu: I did go into more detail in the PR description though

modules/oe_showcase_list_pages/config/install/facets.facet.oelp_oe_sc_news__title.yml
donquixote: I did not see them, but ok.

donquixote: But then it would be "Keywords", not "Keyword"?
I was not able to find an explicit instruction for this in confluence.

donquixote (approved):

escuriola (approved):

PR #160, opened 2022-06-21
OEL-1662: Add event registration url field

escuriola:
### Description

Add event registration url field.

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/tests/src/ExistingSite/ShowcaseExistingSiteCreateEventTest.php
+++ b/tests/src/ExistingSite/ShowcaseExistingSiteCreateEventTest.php
@@ -114,7 +114,7 @@ class ShowcaseExistingSiteCreateEventTest extends ShowcaseExistingSiteTestBase {
     $assert_session->responseContains('image-test.png');
     $assert_session->responseContains('Starter Image test');
     $assert_session->responseContains('Starter Image test alt');
-    $assert_session->responseContains('Register');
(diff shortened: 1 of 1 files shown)


tests/src/ExistingSite/ShowcaseExistingSiteCreateEventTest.php
```diff
@@ -113,6 +114,7 @@ public function testCreateEvent() {
     $assert_session->responseContains('image-test.png');
     $assert_session->responseContains('Starter Image test');
     $assert_session->responseContains('Starter Image test alt');
+    $assert_session->responseContains('Register');
```
drishu: cool, but lets assert also the target blank thing to assert that our formatter is used

drishu (changes requested):

drishu (dismissed):

donquixote (approved):
I think this one is good enough.
We don't override the config, so the tests from whitelabel cover the details.

tests/src/ExistingSite/ShowcaseExistingSiteCreateEventTest.php
donquixote: This was done.

donquixote (dismissed):

drishu (approved):

donquixote (approved):

PR #163, opened 2022-06-29
OEL-1439: Project list pages.

drishu:
### Description

Add search api fields and facets needed for the project content type.

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/modules/oe_showcase_list_pages/tests/src/ExistingSite/ListPagesTest.php
+++ b/modules/oe_showcase_list_pages/tests/src/ExistingSite/ListPagesTest.php
@@ -83,14 +83,14 @@ class ListPagesTest extends ShowcaseExistingSiteTestBase {
     $this->indexItems('oe_list_pages_index');
 
     // Create the News listing page.
-    $list_page = $this->createListPage('oe_sc_news', [
-      'oelp_oe_sc_news__title' => 'oelp_oe_sc_news__title',
-      'oelp_oe_sc_news__oe_publication_date' => 'oelp_oe_sc_news__oe_publication_date',
+    $list_page = $this->createListPage('News list page', 'oe_sc_news', [
+      'oelp_oe_sc_news__title',
+      'oelp_oe_sc_news__oe_publication_date',
     ]);
     $this->drupalGet($list_page->toUrl());
 
     // Assert that only News items are displayed.
-    $this->assertResultsTitle('Results', 12);
+    $this->assertResultsTitle('News list page', 12);
     $this->assertResults([
       'News number 0',
       'News number 1',
@@ -124,7 +124,7 @@ class ListPagesTest extends ShowcaseExistingSiteTestBase {
     $date_input = $filter_form->findField('Date');
     $date_input->setValue('2022-04-04');
     $search_button->click();
-    $this->assertResultsTitle('Results', 8);
+    $this->assertResultsTitle('News list page', 8);
     $this->assertResults([
       'News number 4',
       'News number 5',
@@ -139,7 +139,7 @@ class ListPagesTest extends ShowcaseExistingSiteTestBase {
     // Filter the News results by title.
     $title_input->setValue('News number 8');
     $search_button->click();
-    $this->assertResultsTitle('Results', 1);
+    $this->assertResultsTitle('News list page', 1);
     $this->assertResults([
       'News number 8',
     ]);
@@ -147,7 +147,7 @@ class ListPagesTest extends ShowcaseExistingSiteTestBase {
     // Filter the News results by content.
     $title_input->setValue('This is a News content number 10');
     $search_button->click();
-    $this->assertResultsTitle('Results', 1);
+    $this->assertResultsTitle('News list page', 1);
     $this->assertResults([
       'News number 10',
     ]);
@@ -155,7 +155,7 @@ class ListPagesTest extends ShowcaseExistingSiteTestBase {
     // Filter the News results by introduction.
     $title_input->setValue('This is a News introduction number 11');
     $search_button->click();
-    $this->assertResultsTitle('Results', 1);
+    $this->assertResultsTitle('News list page', 1);
     $this->assertResults([
       'News number 11',
     ]);
@@ -163,14 +163,14 @@ class ListPagesTest extends ShowcaseExistingSiteTestBase {
     // Filter by a mix of introduction and content in shuffled order.
     $title_input->setValue('6 content number introduction');
     $search_button->click();
-    $this->assertResultsTitle('Results', 1);
+    $this->assertResultsTitle('News list page', 1);
     $this->assertResults([
       'News number 6',
     ]);
     // @todo Remove "7" once we have a default order.
     $title_input->setValue('NUMBER 7 this');
     $search_button->click();
-    $this->assertResultsTitle('Results', 1);
+    $this->assertResultsTitle('News list page', 1);
     $this->assertResults([
       'News number 7',
     ]);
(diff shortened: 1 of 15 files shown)


modules/oe_showcase_list_pages/tests/src/ExistingSite/ListPagesTest.php
```diff
@@ -337,4 +470,39 @@ protected function assertElementsTexts(array $expected, array $elements): void {
...
+    ]);
+
+    /** @var \Drupal\emr\Entity\EntityMetaInterface $list_page_entity_meta */
+    $list_page_entity_meta = $list_page->get('emr_entity_metas')->getEntityMeta('oe_list_page');
+    /** @var \Drupal\oe_list_pages\ListPageWrapper $list_page_entity_meta_wrapper */
+    $list_page_entity_meta_wrapper = $list_page_entity_meta->getWrapper();
+    $list_page_entity_meta_wrapper->setSource('node', $bundle);
+    $list_page_entity_meta_wrapper->setConfiguration([
+      'override_exposed_filters' => 1,
+      'exposed_filters' => $exposed_filters,
```
donquixote: I like this method. But let's simplify the `$exposed_filters` parameter.
Let it expect a simple array of strings, and then here we can call `array_combine($exposed_filters, $exposed_filters)`.
```suggestion
      'exposed_filters' => array_combine($exposed_filters, $exposed_filters),
```
See other comments above.

modules/oe_showcase_list_pages/tests/src/ExistingSite/ListPagesTest.php
```diff
@@ -148,66 +139,57 @@ public function testCreateListPages() {
...
-    $page->checkField('emr_plugins_oe_list_page[wrapper][exposed_filters][oelp_oe_sc_event__location]');
-    $page->checkField('emr_plugins_oe_list_page[wrapper][exposed_filters][oelp_oe_sc_event__title]');
-    $page->checkField('emr_plugins_oe_list_page[wrapper][exposed_filters][oelp_oe_sc_event__oe_sc_event_dates]');
-    $page->pressButton('Save');
-
-    $this->drupalGet('node/' . $node->id());
+    $list_page = $this->createListPage('oe_sc_event', [
+      'oelp_oe_sc_event__location' => 'oelp_oe_sc_event__location',
+      'oelp_oe_sc_event__title' => 'oelp_oe_sc_event__title',
+      'oelp_oe_sc_event__oe_sc_event_dates' => 'oelp_oe_sc_event__oe_sc_event_dates',
```
donquixote: Let's simplify the parameter, then here we can do this:
```suggestion
      'oelp_oe_sc_event__location',
      'oelp_oe_sc_event__title',
      'oelp_oe_sc_event__oe_sc_event_dates',
```
See my other comment below.

modules/oe_showcase_list_pages/tests/src/ExistingSite/ListPagesTest.php
```diff
@@ -83,24 +83,15 @@ public function testCreateListPages() {
...
-    $this->drupalGet('node/' . $node->id() . '/edit');
-    $page->checkField('Override default exposed filters');
-    $page->checkField('emr_plugins_oe_list_page[wrapper][exposed_filters][oelp_oe_sc_news__title]');
-    $page->checkField('emr_plugins_oe_list_page[wrapper][exposed_filters][oelp_oe_sc_news__oe_publication_date]');
-    $page->pressButton('Save');
-
-    $this->drupalGet('node/' . $node->id());
+    $list_page = $this->createListPage('oe_sc_news', [
+      'oelp_oe_sc_news__title' => 'oelp_oe_sc_news__title',
+      'oelp_oe_sc_news__oe_publication_date' => 'oelp_oe_sc_news__oe_publication_date',
```
donquixote: Simplify as suggested in other comments below:
```suggestion
      'oelp_oe_sc_news__title',
      'oelp_oe_sc_news__oe_publication_date',
```

modules/oe_showcase_list_pages/tests/src/ExistingSite/ListPagesTest.php
```diff
@@ -83,24 +83,15 @@ public function testCreateListPages() {
...
+    ]);
+    $this->drupalGet($list_page->toUrl());
```
donquixote, line 90: Actually.. now we no longer use the UI to create these list pages. Is this good?
I agree we don't have to do this each time, but at least once, or not?
Why else do we log in as editor?

modules/oe_showcase_list_pages/tests/src/ExistingSite/ListPagesTest.php
```diff
@@ -337,4 +470,39 @@ protected function assertElementsTexts(array $expected, array $elements): void {
...
+   * @param array $exposed_filters
+   *   Facet machine names linked to the bundle's facet source.
+   *
+   * @return \Drupal\node\NodeInterface
+   *   The list page node created.
+   */
+  protected function createListPage(string $bundle, array $exposed_filters): NodeInterface {
+    $list_page = Node::create([
+      'type' => 'oe_list_page',
+      'title' => 'Results',
```
donquixote: I think 'Results' is confusing as a node title here.
```suggestion
      'title' => 'Example list page',
```

modules/oe_showcase_list_pages/tests/src/ExistingSite/ListPagesTest.php
```diff
@@ -255,43 +237,194 @@ public function testCreateListPages() {
...
+    $filter_end->setValue('lt');
+    $dates[1]->setValue($date_plus_10);
+    $search_button->click();
+    $this->assertResultsTitle('Results', 1);
+    $this->assertResults([
+      'Project ongoing',
+    ]);
+
+    // Filter results by status.
+    $filter_form->findButton('Clear filters')->click();
```
donquixote:
```suggestion
    $filter_form->pressButton('Clear filters');
```
This way the IDE does not need to complain about "could be NULL".

modules/oe_showcase_list_pages/tests/src/ExistingSite/ListPagesTest.php
```diff
@@ -255,43 +237,194 @@ public function testCreateListPages() {
...
+    ]);
+
+    // Assert that the filter form for Events exists.
+    $filter_form = $assert_session->elementExists('css', '#oe-list-pages-facets-form');
+    $filter_status = $filter_form->findField('Status');
+    $filter_type = $filter_form->findField('Project type');
+    $filter_budget = $filter_form->findField('Total budget');
+    $filter_start = $filter_form->findField('Start date');
+    $filter_end = $filter_form->findField('End date');
+    $search_button = $filter_form->find('css', '#edit-submit');
```
donquixote, line 351: We forgot one assertNotNull():
```suggestion
    $search_button = $filter_form->find('css', '#edit-submit');
    $this->assertNotNull($filter_status);
```
My IDE complains if a variable could be NULL..

modules/oe_showcase_list_pages/tests/src/ExistingSite/ListPagesTest.php
donquixote: It also throws an exception if the button is not found, which is good.

donquixote (changes requested):

modules/oe_showcase_list_pages/tests/src/ExistingSite/ListPagesTest.php
donquixote: Discuss in chat if needed.

modules/oe_showcase_list_pages/tests/src/ExistingSite/ListPagesTest.php
drishu: If we will ever want to do something like 
```
      'oelp_oe_sc_event__location' => 'oelp_oe_sc_event__location',
      'oelp_oe_sc_event__title' => 0,
      'oelp_oe_sc_event__oe_sc_event_dates' => 'oelp_oe_sc_event__oe_sc_event_dates',
```
we wont be able to, but for our test seems like an edge case, at least at this point, so ok for now, lets see what the future brings.

modules/oe_showcase_list_pages/tests/src/ExistingSite/ListPagesTest.php
drishu: I yielded

modules/oe_showcase_list_pages/tests/src/ExistingSite/ListPagesTest.php
donquixote: Or we could skip that item in that case.
Although, the reason to include it would be to verify that all the options exist.

But for this we could also pass this instead:
```php
      'oelp_oe_sc_event__location' => TRUE,
      'oelp_oe_sc_event__title' => FALSE,
      'oelp_oe_sc_event__oe_sc_event_dates' => TRUE,
```

donquixote (approved):

escuriola (approved):

PR #165, opened 2022-07-04
OEL-1516: Improve the README for oe_showcase_list_pages.

donquixote:


modules/oe_showcase_list_pages/README.md
```diff
@@ -1,37 +1,104 @@
...
+  * Add more fields to prepare for search facets.
+* Click "Save".
+
+### Facets
+
+Create facets that can be added as filters in a list page:
+
+* Visit `admin/config/search/facets`.
+* Click "Add facet" to add facets:
+  * In "Facet source", choose e.g. "List node:oe_sc_news", if the content type machine name is `oe_sc_news`.
```
donquixote, line 92: Note: The labels of these facet sources are weird, and they are different on this "Add facet" page vs the facets overview page. Please confirm and tell me if I am wrong.

modules/oe_showcase_list_pages/README.md
```diff
@@ -1,37 +1,104 @@
...
+### Search index
+
+Create a search index to make specific content types available for use in list pages:
+
+* Visit `config/search/search-api`.
+* Create a search server, unless you want to reuse one that already exists.
+* Click "Add index".
+* Enter an index name, e.g. "List pages index", and review or adjust the machine name.
+  * The index will be reused across content types, so a machine name like `myproject_list_pages_index` is fine.
+  * It is probably a good idea to have dedicated search indices for list pages and for regular site search.
```
drishu: Give instructions, not suggestions, so do this do that

modules/oe_showcase_list_pages/README.md
```diff
@@ -1,37 +1,104 @@
...
+* In "Datasources", choose "Content".
+* In "Bundles", choose the content types that should be indexed.
+* Click "Save and add fields".
+* In the "Fields" tab, add the fields you need:
```
drishu: not the fields you need, rather mention that these fields will be the filters later on

modules/oe_showcase_list_pages/README.md
```diff
@@ -1,37 +1,104 @@
...
+
+* Visit `admin/config/search/facets`.
+* Click "Add facet" to add facets:
+  * In "Facet source", choose e.g. "List node:oe_sc_news", if the content type machine name is `oe_sc_news`.
+  * In "Field", choose one of the fields created earlier.
+  * Enter a name/label, and review the machine name:
+    * Facets exist in a global namespace, so their machine names should be properly prefixed to make them distinguishable and prevent conflicts.
+    * Facets and facet sources for list pages are _not_ reused across content types, so you may need to create multiple facets for the same field. This means the machine name prefix should include the content type.
+  * Click "Save". This should open the "Edit" tab for the new facet.
+  * In "Widget", it is recommended to choose one of the widgets provided by `oe_list_pages`, such as "List pages date", "List pages fulltext" or "List pages multiselect".
```
drishu: not recommended but mandatory

modules/oe_showcase_list_pages/README.md
```diff
@@ -1,37 +1,104 @@
...
+  * The index will be reused across content types, so a machine name like `myproject_list_pages_index` is fine.
+  * It is probably a good idea to have dedicated search indices for list pages and for regular site search.
+* In "Datasources", choose "Content".
+* In "Bundles", choose the content types that should be indexed.
+* Click "Save and add fields".
+* In the "Fields" tab, add the fields you need:
+  * The "Content type" field is required if the index covers multiple content types.
+  * Add more fields to prepare for search facets.
+* Click "Save".
+
```
drishu, line 85: we should mention that the caches might need to be flushed for the new facet source to appear in the facets page, so once you add a new content type to an existing index, a flush is needed to add facets later.

modules/oe_showcase_list_pages/README.md
```diff
@@ -1,37 +1,104 @@
...
+Look at the dependencies sections in `oe_showcase_list_pages.info.yml` and `oe_whitelabel_list_pages.info.yml`.
+
+Composer packages:
+
+- `openeuropa/oe_list_pages`
+  - `drupal/facets` and `drupal/search_api` will be installed as dependencies
+  - Both `drupal/facets` and `openeuropa/oe_list_pages` may need a patch, see "Notes" below.
+- `drupal/extra_field`
+- `openeuropa/oe_whitelabel` - if you want OpenEuropa Whitelabel as a theme.
+
```
drishu, line 47: I'd add here some json to exemplify

modules/oe_showcase_list_pages/README.md
```diff
@@ -1,37 +1,104 @@
...
+Modules:
```
drishu: Enable modules:

drishu (changes requested):
Remove the oe_list_pages patch from the readme, it was merged and released, see https://github.com/openeuropa/oe_list_pages/releases/tag/0.18.0
so maybe in the required modules section specify the min version or in the composer json example I requested.

modules/oe_showcase_list_pages/README.md
donquixote: Not sure what kind of json to add. Snippets that would be added to "require" and "patches"?
I want to avoid adding explicit version numbers here, so that we don't have to update the README all the time.
Maybe I could just refer to the composer.json in showcase?

modules/oe_showcase_list_pages/README.md
donquixote: Not sure about this. Either we use a verb in both cases or never.
E.g. either "Require Composer packages" and "Install Drupal modules" or just "Composer packages" and "Drupal modules".

modules/oe_showcase_list_pages/README.md
donquixote: This section is meant for site builders, who need to figure out what works best for their project.
The sub-points here aim to give some hints about the strategy when planning multiple search indices for a website.

The problem of proper namespacing and reuse of different config entities (search index, facets, fields in a search index) came up when we created this configuration in showcase.

Another way to phrase this:

> Note: In OE Showcase, we decided to use the same search index for all list pages, but use distinct indices for site search or other purposes.

I could give this another shot now but perhaps we should first agree?

modules/oe_showcase_list_pages/README.md
donquixote: Well the fields are needed for facets, and then the facets become filters. Except for the "Content type" field which should always be added for multi content type support.

modules/oe_showcase_list_pages/README.md
donquixote: The other widgets won't work? Good to know.
I just tried and looked into the code, it seems they need to implement `ListPagesWidgetInterface`.

modules/oe_showcase_list_pages/README.md
donquixote: I mentioned cache and indexing process.

donquixote:
> Remove the oe_list_pages patch from the readme, it was merged and released, see https://github.com/openeuropa/oe_list_pages/releases/tag/0.18.0

The patch is still in the "patches" section in composer.json in showcase.
We should remove it there if we want to remove it from the README.

donquixote:
> > Remove the oe_list_pages patch from the readme, it was merged and released, see https://github.com/openeuropa/oe_list_pages/releases/tag/0.18.0
> 
> The patch is still in the "patches" section in composer.json in showcase. We should remove it there if we want to remove it from the README.

Let's fix this first: https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-1818

modules/oe_showcase_list_pages/README.md
donquixote: Let's fix https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-1818 first to get rid of patches.

modules/oe_showcase_list_pages/README.md
drishu: this "Note: We recommend to have one shared search index" is ok

modules/oe_showcase_list_pages/README.md
drishu: ok

modules/oe_showcase_list_pages/README.md
drishu: Its better now

drishu (dismissed):
I don't think we need the conclusions section but lets move on

brummbar (dismissed):
Looks good, the only step missing is to mark the search index to be used in list pages. It's a new option that was added in recent versions.

donquixote:
@brummbar 

> Looks good, the only step missing is to mark the search index to be used in list pages. It's a new option that was added in recent versions.

I added a step for this.
However, the README will now be incorrect until we actually udpate oe_list_pages to 0.18.1.

donquixote:
@drishu 

> I don't think we need the conclusions section but lets move on

I don't like the name "Conclusions" for this, I could not think of anything better.
The purpose of the section is to lead the reader back to "Instructions for content editors".

While writing I thought it a good idea to have "Instructions for content editors" before the "Instructions for site builders", because this illustrates the purpose and motivation of the module, it is the part that will happen repeatedly after initial configuration, and it is also the first thing people should check if they test-run showcase.
But this reversal is against the chronological order in which things would be done in a new project.

drishu (approved):

escuriola (approved):
Approved in behalf of @brummbar

PR #166, opened 2022-07-06
OEL-1553: Person integration

escuriola:
### Description

Person integration

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


tests/src/ExistingSite/ShowcaseExistingSiteCreatePersonTest.php
```diff
@@ -0,0 +1,126 @@
...
+
+use Drupal\file\Entity\File;
+use Drupal\media\Entity\Media;
+use Drupal\Tests\media\Traits\MediaTypeCreationTrait;
+use Drupal\Tests\TestFileCreationTrait;
+
+/**
+ * Class to test Person content type on existing site tests.
+ */
+class ShowcaseExistingSiteCreatePersonTest extends ShowcaseExistingSiteTestBase {
```
brummbar: No need to put the prefix ShowcaseExistingSite, we said we would have to drop this from the other tests too.
And no need to write Create either.

```suggestion
class PersonContentTest extends ShowcaseExistingSiteTestBase {
```

tests/src/ExistingSite/ShowcaseExistingSiteCreatePersonTest.php
```diff
@@ -0,0 +1,126 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function setUp() {
+    parent::setUp();
+
+    // Create user.
+    $user = $this->createUser([]);
+    $this->drupalLogin($user);
```
brummbar: Just move this to the test method.

tests/src/ExistingSite/ShowcaseExistingSiteCreatePersonTest.php
```diff
@@ -0,0 +1,126 @@
...
+        [
+          'target_id' => (int) $document_file->id(),
+          'alt' => 'Event document alt',
+          'title' => 'Event document title',
+        ],
+      ],
+    ]);
+    $media_document->save();
+
+    // Assert editors don't have permissions to create Person items.
```
brummbar: This line is wrong. We are not logged in as editors, and editor MUST be able to edit persons. What are we asserting here?

themes/oe_showcase_theme/oe_showcase_theme.post_update.php
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+declare(strict_types = 1);
+
+use Drupal\block\Entity\Block;
+
+/**
+ * Add person bundle to the social share block.
+ */
+function oe_showcase_theme_post_update_00001(&$sandbox) {
```
brummbar: Post updates for the profile should go in the profile. Themes don't have post updates.

tests/src/ExistingSite/ShowcaseExistingSiteCreatePersonTest.php
```diff
@@ -0,0 +1,126 @@
...
+
+namespace Drupal\Tests\oe_showcase\ExistingSite;
+
+use Drupal\file\Entity\File;
+use Drupal\media\Entity\Media;
+use Drupal\Tests\media\Traits\MediaTypeCreationTrait;
+use Drupal\Tests\TestFileCreationTrait;
+
+/**
+ * Class to test Person content type on existing site tests.
```
brummbar: We need to improve when we can on existing code and comments. There's no value in most of the text of this comment.
```suggestion
 * Tests the person content type.
```

oe_showcase.post_update.php
```diff
@@ -243,3 +243,31 @@ function oe_showcase_post_update_00010(): void {
     ->set('default', 'oe_showcase_theme')
     ->save();
 }
+
+/**
+ * Enable person content type. Grant permissions to editor.
+ */
+function oe_showcase_post_update_00011(): void {
```
brummbar: This is missing the assigning of text formats to text fields. As editor I can access all formats now in all person related fields.

brummbar (changes requested):

tests/src/ExistingSite/ShowcaseExistingSiteCreatePersonTest.php
escuriola: It's a good idea renaming now event and news tests  or should I create a follow up ticket to do?

oe_showcase.post_update.php
```diff
@@ -243,3 +243,64 @@ function oe_showcase_post_update_00010(): void {
...
+  if ($role === NULL) {
+    throw new \Exception("Role not found: 'editor'.");
+  }
+  foreach ($permissions as $permission) {
+    $role->grantPermission($permission);
+  }
+  $role->save();
+
+  // Configure text formats in rich text fields for person.
+  $field_names_by_text_format = [
```
brummbar: It's not by text format anymore.

oe_showcase.post_update.php
```diff
@@ -243,3 +243,64 @@ function oe_showcase_post_update_00010(): void {
...
+    'context_mapping' => [
+      'node' => '@node.node_route_context:node',
+    ],
+    'bundles' => [
+      'oe_sc_event' => 'oe_sc_event',
+      'oe_sc_news' => 'oe_sc_news',
+      'oe_sc_person' => 'oe_sc_person',
+      'oe_showcase_page' => 'oe_showcase_page',
+    ],
+  ]);
```
brummbar: When we have to change one part inside only, we can first get the visibility config, set our missing value and then set it back with setVisibilityConfig().

brummbar (changes requested):
Two minor remarks but it's good.

modules/oe_showcase_list_pages/oe_showcase_list_pages.post_update.php
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+declare(strict_types = 1);
+
+use Drupal\oe_bootstrap_theme\ConfigImporter;
+
+/**
+ * Enable person facets.
+ */
+function oe_showcase_list_pages_post_update_00001(): void {
```
brummbar: I just realised we cannot do this here:
image
By luck we don't have any dependency to the person in these configs, so no errors are thrown, otherwise this update would have failed.
Let's run this post update also in the profile itself.

oe_showcase.post_update.php
```diff
@@ -243,3 +243,64 @@ function oe_showcase_post_update_00010(): void {
     ->set('default', 'oe_showcase_theme')
     ->save();
 }
+
+/**
+ * Install and configure person. Grant permissions to editor.
```
brummbar: Since we are moving the other post update content here, let's limit to "Install and configure person".

oe_showcase.post_update.php
```diff
@@ -243,3 +243,64 @@ function oe_showcase_post_update_00010(): void {
...
+  }
+  foreach ($permissions as $permission) {
+    $role->grantPermission($permission);
+  }
+  $role->save();
+
+  // Configure text formats in rich text fields for person.
+  $field_names_by_text_format = [
+    'oe_sc_person_additional_info',
+    'oe_summary',
```
brummbar: I actually don't think we should use full rich text here. Let's do like other fields and use simple_rich_text.

brummbar (changes requested):

oe_showcase.post_update.php
```diff
@@ -243,3 +243,63 @@ function oe_showcase_post_update_00010(): void {
...
+  ];
+  foreach ($field_names as $field_name) {
+    $field_id = "node.oe_sc_person.$field_name";
+    $field = FieldConfig::load($field_id);
+    if ($field === NULL) {
+      throw new \Exception("Field not found: '$field_id'.");
+    }
+    $field->setThirdPartySetting('allowed_formats', 'allowed_formats', ['simple_rich_text']);
+    $field->save();
+  }
```
donquixote, line 339: Now we set these text formats in post_update, but we don't set them through a field config yml?
It means if you install clean from latest version you don't get this config.

oe_showcase.post_update.php
```diff
@@ -243,3 +243,63 @@ function oe_showcase_post_update_00010(): void {
...
+  ConfigImporter::importMultiple('module', 'oe_showcase_list_pages', '/config/post_updates/00001_person_facets', $configs);
+
+  // Add person bundle to the social share block.
+  $block = Block::load('oe_showcase_theme_social_share');
+  $visibility = $block->getVisibility();
+  if (isset($visibility['entity_bundle:node']['bundles'])) {
+    $visibility['entity_bundle:node']['bundles']['oe_sc_person'] = 'oe_sc_person';
+    $block->setVisibilityConfig('entity_bundle:node', $visibility['entity_bundle:node']);
+    $block->save();
+  }
```
donquixote, line 357: Note:
Interesting asymmetry of methods on the block class: The getter gets the complete visibility settings, but the setter only sets one specific entry.

oe_showcase.post_update.php
```diff
@@ -243,3 +243,63 @@ function oe_showcase_post_update_00010(): void {
...
+    'facets.facet.oelp_oe_sc_person__title',
+    'search_api.index.oe_list_pages_index',
+  ];
+
+  ConfigImporter::importMultiple('module', 'oe_showcase_list_pages', '/config/post_updates/00001_person_facets', $configs);
+
+  // Add person bundle to the social share block.
+  $block = Block::load('oe_showcase_theme_social_share');
+  $visibility = $block->getVisibility();
+  if (isset($visibility['entity_bundle:node']['bundles'])) {
```
donquixote, line 353: We use a condition here to cover the case that no other visibility settings are configured for node types.
But don't cover the case where the entire block placement is not existing (`$block === NULL`), perhaps because it was renamed.

It is all fine because we are in showcase and so we have a known target.
But still inconsistent.

@drishu Thoughts?

donquixote (changes requested):

tests/src/ExistingSite/ShowcaseExistingSiteCreatePersonTest.php
drishu: Lets create a small ticket where we harmonize the test names

tests/src/ExistingSite/PersonTest.php
```diff
@@ -0,0 +1,119 @@
...
+        [
+          'target_id' => (int) $document_file->id(),
+          'alt' => 'Person document alt',
+          'title' => 'Person document title',
+        ],
+      ],
+    ]);
+    $media_document->save();
+
+    // Assert non editor user has no permissions to create Person items.
```
drishu: rather "Assert users without permission to create Person don't have access."

oe_showcase.post_update.php
drishu: Post update is only for existing websites, so oesa for now, and we know oe_showcase_theme_social_share is there.

tests/src/ExistingSite/PersonTest.php
```diff
@@ -0,0 +1,119 @@
...
+    $assert_session->responseContains('Starter Image test alt');
+    $assert_session->pageTextContains('Germany');
+    $assert_session->pageTextContains('DG TEST');
+    $assert_session->pageTextContains('Director');
+    $assert_session->pageTextContains('Linkedin');
+    $assert_session->responseContains('Person document test');
+
+    $this->assertSocialShareBlock();
+  }
+
```
drishu, line 118: This is nice but in the list pages we also need to test the Person

drishu (changes requested):
Failed to execute git checkout 'b79a04ce88c46d7ab4ce047695ed7c3b10dfc279' -- && git reset --hard 'b79a04ce88c46d7ab4ce047695ed7c3b10dfc279' -- 

Check drone, I think composer lock needs to be updated

config/install/field.field.node.oe_sc_person.oe_sc_person_additional_info.yml
```diff
@@ -0,0 +1,25 @@
...
+  config:
+    - field.storage.node.oe_sc_person_additional_info
+    - node.type.oe_sc_person
+  module:
+    - allowed_formats
+    - text
+third_party_settings:
+  allowed_formats:
+    allowed_formats:
+      - rich_text
```
drishu, line 13: I changed it form simple_rich_text

modules/oe_showcase_list_pages/oe_showcase_list_pages.post_update.php
```diff
@@ -10,16 +10,17 @@
 use Drupal\oe_bootstrap_theme\ConfigImporter;
 
 /**
- * Add project filters.
+ * Add project and person filters.
```
drishu, line 13: Had to make this change, the project filters are not released yet, so technically will be ok, reason is
```
DivisionByZeroError: Division by zero in Drupal\search_api\Task\TaskManager  
  ->processBatch() (line 370 of /test/oe_showcase/build/modules/contrib/searc  
  h_api/src/Task/TaskManager.php).
```

drishu (dismissed):

donquixote (approved):

brummbar (approved):

PR #170, opened 2022-07-13
OEL-1761: Create administrator role

escuriola:
### Description

Create new administrator admin role.

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/config/install/user.role.administrator.yml
+++ b/config/install/user.role.administrator.yml
@@ -1,0 +1,8 @@
+langcode: en
+status: true
+dependencies: {  }
+id: administrator
+label: Administrator
+weight: 5
+is_admin: true
+permissions: {  }
--- a/oe_showcase.info.yml
+++ b/oe_showcase.info.yml
@@ -112,6 +112,7 @@ config_devel:
     - purge_users.settings
     - roleassign.settings
     - system.theme
+    - user.role.administrator
     - user.role.anonymous
     - user.role.authenticated
     - user.role.configure_page_feedback_form
--- a/tests/src/ExistingSite/AuthorisationTest.php
+++ b/tests/src/ExistingSite/AuthorisationTest.php
@@ -92,6 +92,28 @@ class AuthorisationTest extends ShowcaseExistingSiteTestBase {
   }
 
   /**
+   * Authenticated user cannot access restricted pages.
+   */
+  public function testAdministatorAccess(): void {
+    $this->drupalLogin($this->createUserWithRoles(['administrator']));
+    $paths = [
+      '/admin',
+      '/admin/config',
+      '/admin/reports/status',
+      '/admin/content',
+      '/admin/content/media',
+      '/admin/people',
+      '/admin/structure',
+      '/node/add',
+    ];
+
+    foreach ($paths as $path) {
+      $this->drupalGet($path);
+      $this->assertSession()->statusCodeEquals(200);
+    }
+  }
+
+  /**
    * Manage users user cannot access restricted pages.
    */
   public function testManageUsersAccess(): void {
--- a/tests/src/ExistingSite/AuthorisationTest.php
+++ b/tests/src/ExistingSite/AuthorisationTest.php
@@ -94,17 +94,39 @@ class AuthorisationTest extends ShowcaseExistingSiteTestBase {
   /**
    * Authenticated user cannot access restricted pages.
    */
-  public function testAdministatorAccess(): void {
+  public function testAdministratorAccess(): void {
     $this->drupalLogin($this->createUserWithRoles(['administrator']));
     $paths = [
       '/admin',
+      '/admin/structure/block',
+      '/admin/structure/block/manage/oe_showcase_theme_main_page_content',
+      '/admin/structure/contact',
+      '/admin/structure/contact/manage/example_contact_form',
+      '/admin/structure/oe_contact_type',
+      '/admin/structure/types',
+      '/admin/structure/types/manage/oe_showcase_page',
+      '/admin/structure/menu',
+      '/admin/structure/menu/manage/main',
+      '/admin/structure/paragraphs_type',
+      '/admin/structure/paragraphs_type/oe_facts_figures',
+      '/admin/structure/taxonomy',
+      '/admin/appearance',
+      '/admin/appearance/settings/oe_showcase_theme',
+      '/admin/modules',
       '/admin/config',
+      '/admin/config/system/site-information',
+      '/admin/config/development/configuration',
+      '/admin/config/development/maintenance',
+      '/admin/config/regional/settings',
+      '/admin/config/regional/language',
+      '/admin/people',
+      '/admin/people/permissions',
+      'admin/people/roles',
       '/admin/reports/status',
       '/admin/content',
       '/admin/content/media',
-      '/admin/people',
-      '/admin/structure',
       '/node/add',
+      '/node/add/oe_showcase_page',
     ];
 
     foreach ($paths as $path) {
--- a/tests/src/ExistingSite/AuthorisationTest.php
+++ b/tests/src/ExistingSite/AuthorisationTest.php
@@ -92,39 +92,39 @@ class AuthorisationTest extends ShowcaseExistingSiteTestBase {
   }
 
   /**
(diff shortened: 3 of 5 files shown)


drishu (changes requested):
Looks good, just one more thing, at a minimum we should assert in AuthorisationTest that a user with this role can access the stuff that only an admin should/could

tests/src/ExistingSite/AuthorisationTest.php
```diff
@@ -91,6 +91,28 @@ public function testAuthenticatedAccess(): void {
...
+    $this->drupalLogin($this->createUserWithRoles(['administrator']));
+    $paths = [
+      '/admin',
+      '/admin/config',
+      '/admin/reports/status',
+      '/admin/content',
+      '/admin/content/media',
+      '/admin/people',
+      '/admin/structure',
+      '/node/add',
```
drishu, line 128: we need more stuff here like the permissions page, the roles page, managing entities like content types, structure blocks, block config, etc.. all the pages we need to have access to when we assert a deployment.

drishu (changes requested):

tests/src/ExistingSite/AuthorisationTest.php
```diff
@@ -94,17 +94,39 @@ public function testAuthenticatedAccess(): void {
...
+      '/admin/modules',
       '/admin/config',
+      '/admin/config/system/site-information',
+      '/admin/config/development/configuration',
+      '/admin/config/development/maintenance',
+      '/admin/config/regional/settings',
+      '/admin/config/regional/language',
+      '/admin/people',
+      '/admin/people/permissions',
+      'admin/people/roles',
```
drishu: a slash is missing here, but its fine, fix only if next reviewer has more feedback

drishu (dismissed):

tests/src/ExistingSite/AuthorisationTest.php
```diff
@@ -91,6 +91,50 @@ public function testAuthenticatedAccess(): void {
     }
   }
 
+  /**
+   * Authenticated user cannot access restricted pages.
```
donquixote: Currently this comment does not describe what the test does.
We are only testing that the admin role _can_ access these pages, not that authenticated role _cannot_.

### Option 1:
Only change the comment to describe what the method currently does.
```suggestion
   * Tests that the administrator can see restricted administration pages.
```

### Option 2:
Add checks to verify that authenticated and anonymous cannot see these pages.
Modify the comment accordingly:
```suggestion
   * The administrator can see pages that authenticated users or anonymous cannot.
```

### Option 3:
Modify the code to check which roles can access which of the admin pages.
E.g. `/admin` can be seen by 'administrator' and by 4 other roles.
Modify the comment accordingly.
This would require to further modify the methods in `AssertPathAccessTrait`, more than I am comfortable to ask for in this PR.

Imo let's go for option 2.

tests/src/ExistingSite/AuthorisationTest.php
```diff
@@ -91,6 +91,50 @@ public function testAuthenticatedAccess(): void {
...
+   */
+  public function testAdministratorAccess(): void {
+    $this->drupalLogin($this->createUserWithRoles(['administrator']));
+    $paths = [
+      '/admin',
```
donquixote, line 100: Just as an observation, there are 4 other roles that can see `'/admin'`:
- 'configure_page_feedback_form', 
- 'manage_contact_forms'
- 'editor'
- 'manage_users'.

Other paths are accessible by fewer roles.

But let's ignore this for now, we might do a follow-up to improve these tests.

config/install/user.role.administrator.yml
```diff
@@ -0,0 +1,8 @@
+langcode: en
+status: true
+dependencies: {  }
+id: administrator
+label: Administrator
+weight: 5
+is_admin: true
+permissions: {  }
```
donquixote, line 8: We should add the entry in oe_showcase.info.yml `config_devel/install`.

tests/src/ExistingSite/AuthorisationTest.php
donquixote: Pls add the slash, it hurts to see :)

tests/src/ExistingSite/AuthorisationTest.php
```diff
@@ -91,6 +91,50 @@ public function testAuthenticatedAccess(): void {
...
+      '/admin/config/regional/settings',
+      '/admin/config/regional/language',
+      '/admin/people',
+      '/admin/people/permissions',
+      'admin/people/roles',
+      '/admin/reports/status',
+      '/admin/content',
+      '/admin/content/media',
+      '/node/add',
+      '/node/add/oe_showcase_page',
```
donquixote, line 129: Please sort this list alphabetically.

tests/src/ExistingSite/AuthorisationTest.php
```diff
@@ -91,6 +91,50 @@ public function testAuthenticatedAccess(): void {
...
+    ];
+
+    foreach ($paths as $path) {
+      $this->drupalGet($path);
+      $this->assertSession()->statusCodeEquals(200);
+    }
```
donquixote, line 135: If we go for option 2, we would now check that anonymous and authenticated cannot see these paths.
We can use the `->assertPathsResponseCode()` to avoid repeating the loop, and to have more specific failure messages.

tests/src/ExistingSite/AuthorisationTest.php
donquixote: @drishu Disagree?

donquixote (changes requested):

tests/src/Traits/AssertPathAccessTrait.php
```diff
@@ -77,7 +80,7 @@ protected function assertPathsAccessForbidden(array $paths): void {
 
     // Test all existing roles.
     $all_roles = user_role_names(TRUE);
-    unset($all_roles['authenticated']);
+    unset($all_roles['authenticated'], $all_roles['administrator']);
```
donquixote, line 83: In a follow-up, we could also test that administrator can see these paths.
But let's skip it for now.

tests/src/ExistingSite/AuthorisationTest.php
drishu: option 1

config/install/user.role.administrator.yml
drishu: good catch

tests/src/ExistingSite/AuthorisationTest.php
donquixote: So.. @drishu seems to be ok with option 1.
In that case, just change the comment.

escuriola: I chose option 1 for now after reading the chat discussion. We can create a follow-up ticket if you want to improve it @donquixote.

tests/src/ExistingSite/AuthorisationTest.php
escuriola: Not implemented right now, as you said, we can create a follow-up if we want to improve the test

tests/src/ExistingSite/AuthorisationTest.php
escuriola: Not implemented because option 1 is chosen

tests/src/Traits/AssertPathAccessTrait.php
escuriola: Ok, not implemented.

oe_showcase.info.yml
```diff
@@ -113,6 +113,7 @@ config_devel:
     - roleassign.settings
     - system.theme
     - user.role.anonymous
+    - user.role.administrator
```
donquixote: Alphabetic order!
```suggestion
    - user.role.administrator
    - user.role.anonymous
```

donquixote (approved):

donquixote:
Ok for me now.
The one comment from @drishu I left open. I assume that now we have enough paths, but maybe we want more?
For my taste we should go with what we have now.

drishu (approved):

PR #171, opened 2022-07-14
OEL-1775: Enable social media follow in basic page.

escuriola:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/modules/oe_showcase_page/oe_showcase_page.post_update.php
+++ b/modules/oe_showcase_page/oe_showcase_page.post_update.php
@@ -1,0 +1,22 @@
+<?php
+
+/**
+ * @file
+ * OpenEuropa Showcase Page post updates.
+ */
+
+declare(strict_types = 1);
+
+use Drupal\field\Entity\FieldConfig;
+
+/**
+ * Enable the social_media paragraph.
+ */
+function oe_showcase_page_post_update_00001(&$sandbox): void {
+  $field = FieldConfig::load('node.oe_showcase_page.field_body');
+  $settings = $field->getSetting('handler_settings');
+  $settings['target_bundles']['oe_social_media_follow'] = 'oe_social_media_follow';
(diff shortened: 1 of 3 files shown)


modules/oe_showcase_page/oe_showcase_page.post_update.php
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+declare(strict_types = 1);
+
+use Drupal\field\Entity\FieldConfig;
+
+/**
+ * Enable the social_media paragraph.
+ */
+function oe_showcase_page_post_update_00001(&$sandbox) {
```
drishu:
```suggestion
function oe_showcase_page_post_update_00001(&$sandbox): void {
```
we dont specify return type only if its mixed, and only as long as we support php smaller than 8

drishu (changes requested):

drishu (dismissed):

modules/oe_showcase_page/oe_showcase_page.post_update.php
```diff
@@ -0,0 +1,22 @@
...
+
+use Drupal\field\Entity\FieldConfig;
+
+/**
+ * Enable the social_media paragraph.
+ */
+function oe_showcase_page_post_update_00001(&$sandbox): void {
+  $field = FieldConfig::load('node.oe_showcase_page.field_body');
+  $settings = $field->getSetting('handler_settings');
+  $settings['target_bundles']['oe_social_media_follow'] = 'oe_social_media_follow';
```
donquixote, line 18: We could sort these to avoid noise later on.
But I don't think we need to. Let's ignore.

donquixote (dismissed):

drishu (approved):

donquixote (approved):

PR #178, opened 2022-08-03
OEL-1875: Allow to embed medias in the rich text editor.

brummbar:
Only AV Portal images and videos, remote videos, local images and documents.


tests/src/Traits/MediaCreationTrait.php
```diff
@@ -0,0 +1,284 @@
...
+namespace Drupal\Tests\oe_showcase\Traits;
+
+use Drupal\Component\Utility\NestedArray;
+use Drupal\file\Entity\File;
+use Drupal\media\MediaInterface;
+use weitzman\DrupalTestTraits\Entity\MediaCreationTrait as WeitzmanMediaCreationTrait;
+
+/**
+ * Contains methods to create media entities for testing.
+ *
```
drishu, line 14: would be a nice to have to write here that this can be used only withing ExistingSiteTests

tests/src/ExistingSiteJavascript/WysiwygEmbedTest.php
```diff
@@ -0,0 +1,215 @@
...
+    $this->getSession()->switchToIFrame();
+    $assert_session->assertWaitOnAjaxRequest();
+    $modal_button_pane = $assert_session->elementExists('css', 'div.ui-dialog-buttonpane');
+    $modal_button_pane->findButton('Embed')->press();
+    $assert_session->assertWaitOnAjaxRequest();
+
+    $this->getSession()->switchToIFrame('ckeditor');
+    $assert_session->linkExists($label);
+
+    $this->getSession()->switchToIFrame();
```
drishu, line 178: Very nice work here, the only thing that concerns me is that the av portal part we will need to mocked in near future

drishu (dismissed):

tests/src/ExistingSiteJavascript/WysiwygEmbedTest.php
brummbar: It's already mocked :) https://github.com/openeuropa/oe_showcase/blob/1.x/.drone.yml#L117

drishu: I see now, thx

drishu (approved):

PR #179, opened 2022-08-08
OEL-1747: Remove misleading field description for facts and figures paragraph.

brummbar:


composer.json
```diff
@@ -125,6 +125,9 @@
             "openeuropa/oe_bootstrap_theme": {
                 "latest": "https://github.com/openeuropa/oe_bootstrap_theme/compare/0.1.202208041835...1.x.diff"
             },
+            "openeuropa/oe_paragraphs": {
```
drishu, line 119: I see the change in lock but if we patch against 1.14 it I think its good to bump up oe_paragraphs also in composer.json

drishu (approved):

PR #186, opened 2022-08-30
OEL-1914: Integrate the carousel paragraph.

drishu:


config/post_updates/00019_carousel/field.field.node.oe_showcase_page.field_body.yml
```diff
@@ -0,0 +1,135 @@
...
+      oe_map: oe_map
+      oe_quote: oe_quote
+      oe_rich_text: oe_rich_text
+      oe_social_feed: oe_social_feed
+      oe_social_media_follow: oe_social_media_follow
+      oe_text_feature_media: oe_text_feature_media
+      oe_timeline: oe_timeline
+    negate: 0
+    target_bundles_drag_drop:
+      contact_form:
```
brummbar, line 66: Let's keep these entries sorted alphabetically, it looks like carousel is the only one is not in the dropdown.

modules/oe_showcase_page/tests/src/ExistingSite/PageTest.php
```diff
@@ -326,6 +330,53 @@ public function testCreateShowCasePage() {
       'linkedin',
     );
 
+    // Add Carousel paragraph.
+    $media_1 = $this->createImageMedia(['name' => 'First image']);
+    $media_2 = $this->createImageMedia(['name' => 'Second image']);
```
brummbar: It's already tested, but why not use an avportal image here.

modules/oe_showcase_page/tests/src/ExistingSite/PageTest.php
```diff
@@ -326,6 +330,53 @@ public function testCreateShowCasePage() {
...
+
+    $page->pressButton('Add Carousel');
+    $page->fillField(
+      'field_body[8][subform][field_oe_carousel_items][0][subform][field_oe_title][0][value]',
```
brummbar, line 340: We are never testing that the title is optional.

modules/oe_showcase_page/tests/src/ExistingSite/PageTest.php
```diff
@@ -392,6 +443,24 @@ public function testCreateShowCasePage() {
     $assert_session->pageTextContains('Social share links');
     $assert_session->pageTextContains('Linkedin profile');
 
+    // Assert Carousel.
+    $assert_session->pageTextContains('Carousel item 1');
```
brummbar: We should use the pattern assertion here too.

brummbar (changes requested):
I've noticed one problem with the bcl carousel template. No caption title will be rendered if no caption is entered. We need to check where this should be fixed. If we should make the caption mandatory on our side, or fix the template, or add some states to make it required when other fields are filled.

modules/oe_showcase_page/tests/src/ExistingSite/PageTest.php
drishu: implemented for the second item

brummbar (approved):

PR #187, opened 2022-09-19
OEL-2006: Glossary.

brummbar:


themes/oe_showcase_theme/templates/views/views-view-summary--glossary-page--attachment-1.html.twig
```diff
@@ -0,0 +1,35 @@
...
+ *   This contains:
+ *   - count: A flag indicating whether the count should be displayed.
+ *
+ * @see template_preprocess_views_view_summary()
+ *
+ * @ingroup themeable
+ */
+#}
+{% set _items = [] %}
+{% for row in rows %}
```
drishu, line 24: Lets create a ticket or discuss in the daily if we need to pad

modules/oe_showcase_glossary/src/GlossaryViewHooks.php
```diff
@@ -0,0 +1,270 @@
...
+      return;
+    }
+
+    // Make sure the letter is lowercased, to match the configuration of our
+    // contextual filter settings.
+    $args[] = mb_strtolower($summary_results[0]->name_truncated);
+  }
+
+  /**
+   * Implements hook_views_query_alter().
```
drishu, line 126: This bothers me a bit, even though it is used in the hook_views_query_alter it does not actually implement it, I'd rather have a short desc of what its for and maybe reference the hook in a @see.

modules/oe_showcase_glossary/src/GlossaryViewHooks.php
```diff
@@ -0,0 +1,270 @@
...
+    $tokens = $view->getDisplay()->getArgumentsTokens();
+    $name_argument = $tokens['{{ arguments.name }}'] ?? NULL;
+    $variables['character'] = $name_argument;
+    $variables['total_results'] = $view->total_rows;
+  }
+
+  /**
+   * Preprocess for the summary display.
+   *
+   * Sort the glossary summary entries so that numbers are shown after letters.
```
drishu: Isn't this rather setting the active element when the default is used but not present in the url? and than this comment is more valid in queryAlter

modules/oe_showcase_glossary/tests/src/ExistingSite/GlossaryTest.php
```diff
@@ -0,0 +1,322 @@
...
+
+      $terms_by_number += [$starting_number => []];
+      $terms_by_number[$starting_number][] = $this->createTerm($glossary_vocabulary, [
+        'name' => $starting_number . $this->randomMachineName(),
+        'langcode' => 'en',
+      ]);
+    }
+
+    // Sort the terms alphabetically.
+    $this->sortTerms($terms_by_letter, 'letter');
```
drishu: the second parameter seems like a leftover

themes/oe_showcase_theme/templates/views/views-view--glossary-page--page-1.html.twig
```diff
@@ -0,0 +1,84 @@
...
+ * - total_results: The total number of results.
+ *
+ * @see template_preprocess_views_view()
+ *
+ * @ingroup themeable
+ */
+#}
+<div class="glossary-view">
+  {{ title_prefix }}
+  {{ pattern('content_banner', {
```
drishu, line 41: we need to address on BCL level those classes that are forced that create in this situation extra spacing, we should not assume that this content banner will always be used above the "main" section

themes/oe_showcase_theme/templates/taxonomy/taxonomy-term--glossary--full.html.twig
```diff
@@ -0,0 +1,41 @@
...
+ * - view_mode: View mode, e.g. 'full', 'teaser', etc.
+ *
+ * @see template_preprocess_taxonomy_term()
+ *
+ * @ingroup themeable
+ */
+#}
+<div{{ attributes }}>
+  {{ title_prefix }}
+  <h1 class="fs-1 pb-4 pb-md-4-75 pt-4 pt-md-4-75">{{ name }}</h1>
```
drishu, line 30: We will follow up with supporting more entity types in the content banner block.

drishu (changes requested):

themes/oe_showcase_theme/templates/views/views-view--glossary-page--page-1.html.twig
```diff
@@ -0,0 +1,92 @@
...
+        }) }}
+      </h2>
+    {% endif %}
+
+    {% if rows -%}
+      {{ rows }}
+    {% elseif empty -%}
+      {{ empty }}
+    {% else %}
+      <p>{{ 'No results found.'|t }}</p>
```
drishu: I see now why this message is better but it should go in the config of the empty text, see the failing test.

drishu (dismissed):

PR #193, opened 2022-10-03
OEL-2074: Glossary term permissions and canonical page fixes

brummbar:


oe_showcase.profile
```diff
@@ -18,6 +19,24 @@ function oe_showcase_install_tasks_alter(&$tasks, $install_state): void {
...
+  $tasks['oe_showcase_disable_taxonomy_term_view'] = [];
+}
+
+/**
+ * Disables the taxonomy term canonical page override provided by Views.
+ *
+ * @param array $install_state
+ *   An array of information about the current installation state.
+ */
+function oe_showcase_disable_taxonomy_term_view(array &$install_state): void {
```
drishu, line 32: Since this is a profile I agree that this is the best way to solve it, but, we also needs this in oe_showcase_post_update_00020 so that OESA benefits from it

modules/oe_showcase_glossary/README.md
```diff
@@ -54,6 +54,10 @@ Extra taxonomy term template suggestions are necessary to style the teaser view
 Various hooks implementations are located in [`oe_showcase_glossary.module`](./oe_showcase_glossary.module) and
 implemented in a dedicated class, [`GlossaryViewHooks`](./src/GlossaryViewHooks.php).
 
+An install task [`oe_showcase_disable_taxonomy_term_view()`](../../oe_showcase.profile) was implemented to disable the
```
drishu, line 57: I feel this is a "private matter" of our profile, and would rather point them to our update hook so they can see what else they need to pay attention to besides the module code.
And there, in the post_update, maybe even leave 1 or 2 comments, saying "this is what you need to do for your glossary module, your roles, but via drush cex".. something of sorts.

modules/oe_showcase_glossary/tests/src/ExistingSite/GlossaryTest.php
```diff
@@ -137,6 +141,36 @@ public function testGlossaryView(): void {
...
+   * Tests that the editor role can manage the glossary vocabulary.
+   */
+  public function testGlossaryVocabularyAccess(): void {
+    $glossary_vocabulary = Vocabulary::load('glossary');
+    $term = $this->createTerm($glossary_vocabulary);
+
+    $paths = [
+      '/admin/structure/taxonomy/manage/glossary/overview',
+      '/admin/structure/taxonomy/manage/glossary/add',
+      $term->toUrl('edit-form')->setAbsolute()->toString(),
```
drishu, line 167: like what you did here :)

drishu (changes requested):

oe_showcase.profile
brummbar: Opsie

modules/oe_showcase_glossary/README.md
brummbar: In a normal website there won't be any post updates hook to update the configuration, but they should rather do it via configuration export as it's indicated. We shouldn't drive them towards update hooks.

brummbar: We could remove this completely and just leave the reference in the "How to install" steps. Permissions is also something out of scope of this readme, so we should just stick to the "disable the view" part.

drishu: That's what I meant with the drush cex, not to steer them towards update hooks, but ok, I guess it wont bring them added value

drishu (approved):

PR #195, opened 2022-10-13
OEL-1414: Integrate OE List Page sort.

drishu:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/modules/oe_showcase_list_pages/src/EventSubscriber/ListPagesSubscriber.php
+++ b/modules/oe_showcase_list_pages/src/EventSubscriber/ListPagesSubscriber.php
@@ -1,0 +1,84 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\oe_showcase_list_pages\EventSubscriber;
+
+use Drupal\oe_list_pages\ListPageEvents;
+use Drupal\oe_list_pages\ListPageSortAlterEvent;
+use Drupal\oe_list_pages\ListPageSortOptionsResolver;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+/**
+ * OpenEuropa Showcase List Pages event subscriber.
+ */
+class ListPagesSubscriber implements EventSubscriberInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function getSubscribedEvents() {
+    return [
+      ListPageEvents::ALTER_SORT_OPTIONS => ['onSortOptionsAlter'],
+    ];
+  }
+
+  /**
+   * Event handler for altering the sort options.
+   *
+   * @param \Drupal\oe_list_pages\ListPageSortAlterEvent $event
+   *   The event.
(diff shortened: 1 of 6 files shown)


modules/oe_showcase_list_pages/src/EventSubscriber/ListPagesSubscriber.php
```diff
@@ -0,0 +1,84 @@
...
+      ListPageSortOptionsResolver::SCOPE_USER => [
+        'node' => [
+          'oe_sc_news' => [
+            'oe_publication_date__ASC' => 'Date ASC',
+            'oe_publication_date__DESC' => 'Date DESC',
+            'title__ASC' => 'A-Z',
+            'title__DESC' => 'Z-A',
+          ],
+          'oe_sc_event' => [
+            'oe_sc_event_dates' => 'Date',
```
brummbar: Is the sorting direction omitted on purpose in here? Because the code doesn't seem to account for it. It works, but it gets into the url as null.

brummbar (changes requested):

modules/oe_showcase_list_pages/tests/src/ExistingSiteJavascript/ListPagesTest.php
```diff
@@ -2,33 +2,47 @@
...
-    $this->drupalLogin($user);
+    $this->editorUser = $this->createUser([]);
+    $this->editorUser->addRole('editor');
+    $this->editorUser->save();
+
+    // Prevent toolbar from overlapping.
+    // Maximizing the window gives inconsistent results, so we remove the
+    // element all together.
+    $role = Role::load('editor');
+    $role->revokePermission('access toolbar');
```
drishu, line 44: Heads up, the maximize window was not cutting it anymore, I've also tried hiding the toolbar via js, same result, element overlap, so I had to do this.

modules/oe_showcase_list_pages/tests/src/ExistingSiteJavascript/ListPagesTest.php
```diff
@@ -402,73 +512,113 @@ public function testCreateListPages() {
...
+    $this->scrollIntoView('#edit-submit');
     $search_button->click();
-    $this->assertResultsTitle('Project list page', 2);
+    $this->assertResultsTitle('Project List Page', 2);
+    $this->assertResults([
+      'Project ongoing',
+      'Project pending',
+    ]);
+
+    $this->scrollIntoView('#edit-reset');
```
drishu, line 606: I've added these extra test cases, and noticed that the order for the budget is actually sorted as string, not as number, can be fixed only when we have a "number" widget in oe_list_pages (as far as I can see)

brummbar (approved):

PR #196, opened 2022-10-14
OEL-1990: Fix allowed formats in formatted text fields.

drishu:
### Description

The format for certain formatted text fields was not set so the wrong text format ended up in existing content.
To fix this we need to:

- Set the correct text formats in the field configuration
- Add the correct allowed formats for existing content
- Validate for the correct format to be used by editors
- Disable the full_html format as its usage is risky

### Change log

- Added: 
  - Field config overrides
  - Widget alters to inject correct format
  - Format validation
  - Upgrade path


tests/src/Traits/WysiwygTrait.php
```diff
@@ -0,0 +1,46 @@
...
+  /**
+   * Asserts a field has the expected WYSIWYG editor.
+   *
+   * @param \Behat\Mink\Element\NodeElement $field
+   *   The field to check.
+   * @param string $format
+   *   The format.
+   */
+  protected function assertFieldHasWysiwyg(NodeElement $field, string $format): void {
+    $parent = $this->getParentByAttribute($field, 'class', 'text-format-wrapper');
```
drishu: I've tried all sorts of xpath but it's not finding my element wrapper

tests/src/Traits/TraversingTrait.php
```diff
@@ -30,4 +30,35 @@ protected function getSelectOptions(NodeElement $select): array {
...
+   *
+   * @return \Behat\Mink\Element\NodeElement|void
+   *   The parent element if found.
+   */
+  protected function getParentByAttribute(NodeElement $element, string $attribute, string $value, int $depth = 5) {
+    $parent = $element->getParent();
+
+    // Limit the depth to prevent infinite loop.
+    for ($i = 0; $i < $depth; $i++) {
+      if (!$parent instanceof NodeElement) {
```
drishu: just in case, but from my tests it doesn't happen, instead it would go into an infinite loop in case I use a while loop

oe_showcase.post_update.php
```diff
@@ -495,3 +495,29 @@ function oe_showcase_post_update_00020(): void {
...
+    'user.user.field_bio',
+  ];
+
+  foreach ($fields as $field_name) {
+    /** @var \Drupal\Core\Field\FieldConfigInterface $field_instance */
+    $field_instance = $field_storage->load($field_name);
+    if (!$field_instance) {
+      continue;
+    }
+    $field_instance->setThirdPartySetting('allowed_formats', 'allowed_formats', ['rich_text']);
```
brummbar: This is not as straightforward as it seems. When you have existing content, this happens:

image

What we need to follow a similar approach to this PR https://github.com/ec-europa/ewcms/pull/805/files#diff-87cdcb75ea06e72dd3577cd72f428924eba5569fd8f50d86438a0ebc69de450eR4481
Basically:

- the config contains only the allowed values;
- for updated websites, we allow both formats for existing content;
- when an existing content is updated, we add validation to enforce changing to the correct format.

I think we can drop the part with the install task that makes fresh and existing installs in the same state.

tests/src/Traits/TraversingTrait.php
```diff
@@ -30,4 +30,35 @@ protected function getSelectOptions(NodeElement $select): array {
...
+   *   The parent attribute for comparison.
+   * @param string $value
+   *   The parent attribute value to compare to.
+   * @param int $depth
+   *   The maximum depth to limit to.
+   *
+   * @return \Behat\Mink\Element\NodeElement|void
+   *   The parent element if found.
+   */
+  protected function getParentByAttribute(NodeElement $element, string $attribute, string $value, int $depth = 5) {
```
brummbar: I would drop this method, we can use an xpath to get a parent.
`$field->find('xpath', '/ancestor::div[@class and contains(concat(" ", normalize-space(@class), " "), " text-format-wrapper ")]')`

tests/src/Traits/WysiwygTrait.php
```diff
@@ -0,0 +1,46 @@
...
+
+  /**
+   * Asserts a field has the expected text format.
+   *
+   * @param \Behat\Mink\Element\NodeElement $field
+   *   The field to check.
+   * @param string $format
+   *   The format.
+   */
+  protected function assertFieldHasFormat(NodeElement $field, string $format): void {
```
brummbar: So the difference between these two functions is that one asserts that a format is set, while the other asserts the format description, correct?
Unfortunately from the name it's not very clear.
Let's use [oe_editorial](https://github.com/openeuropa/oe_editorial/blob/master/oe_editorial.module#L27) so that we have the link changed to reflect which format we have. Then we can assert on that one.
I would remove these methods for now and just have one method which returns the format used for a field.
Look at what we can reuse from wysiwyg trait in ewpp. Also let's keep out assertions from traits without assert in the name. Better to just return the format.

tests/src/ExistingSite/PersonTest.php
```diff
@@ -86,12 +88,16 @@ public function testCreatePerson() {
...
+    $field = $page->findField('Short description');
+    $this->assertFieldHasFormat($field, 'simple_rich_text');
+    $field->setValue('Example short description field.');
     $page->fillField('Use existing media', $media_image->getName() . ' (' . $media_image->id() . ')');
     $page->fillField('Country', 'DE');
     $page->fillField('Occupation', 'DG TEST');
     $page->fillField('Position', 'Director');
-    $page->fillField('Additional information', 'Example additional information field.');
+    $field = $page->findField('Additional information');
+    $this->assertFieldHasWysiwyg($field, 'rich_text');
```
brummbar: Basically with the comments below/above this becomes
```suggestion
    $this->assertEquals('rich_text', $this->getWysiwigTextFormat($field));
```

brummbar (changes requested):

tests/src/Traits/WysiwygTrait.php
drishu: the ewcms wysiwyg trait doesn't help us here, this is how its done there https://github.com/ec-europa/ewcms/blob/4e6baee6b8cc859ab16e3beaaaf436e44e047b9f/tests/Behat/DrupalContext.php#L2351, found this after you pointed me to the oe_editorial alter

oe_showcase.post_update.php
drishu: "the config contains only the allowed values" if you mean only rich text without the full_html then I dont like it cuz the update should match the install. Regarding the install task I agreed we dont need it, having the same allowed formats in both install and update takes care of that.

oe_showcase.profile
```diff
@@ -145,3 +148,170 @@ function oe_showcase_field_widget_entity_browser_entity_reference_form_alter(&$e
...
+    return;
+  }
+
+  $format = FilterFormat::load($expected_format);
+  $path[] = 'format';
+  $form_state->setErrorByName(implode('][', $path), t('The field %field_name supports only the %format format.', [
+    '%field_name' => $element['#title'],
+    '%format' => $format->label(),
+  ]));
+}
```
brummbar, line 273: We are missing one scenario which is the text field for timeline items. It's a custom one so we need a hook

```
/**
 * Implements hook_field_widget_WIDGET_TYPE_form_alter().
 */
function oe_showcase_field_widget_timeline_widget_form_alter(&$element, FormStateInterface $form_state, $context) {
  // Bail out if it's a default value widget.
  if ($context['default']) {
    return;
  }

  if (!isset($element['body']['#format'])) {
    return;
  }
  // Force Simple rich text format for timeline body.
  $element['body']['#format'] = 'rich_text';
}
```

oe_showcase.post_update.php
```diff
@@ -535,3 +536,63 @@ function oe_showcase_post_update_00021(): void {
   $editor->grantPermission('access images_and_videos entity browser pages');
   $editor->save();
 }
+
+/**
+ * Fix allowed formats in formatted text fields.
+ *
+ * @see https://github.com/ec-europa/ewcms/blob/develop/post_updates/ewcms_post_update_200.inc#L2096
```
brummbar: Let's move this link from here as it generates a bit of noise when running updb.

brummbar (changes requested):

brummbar (approved):

PR #201, opened 2022-11-11
OEL-2107: Added styles for gallery modal.

tibi2303:
### Description

Improved images styles displayed in the gallery modal.

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/oe_showcase.post_update.php
+++ b/oe_showcase.post_update.php
@@ -535,3 +535,10 @@ function oe_showcase_post_update_00021(): void {
   $editor->grantPermission('access images_and_videos entity browser pages');
   $editor->save();
 }
+
+/**
+ * Update the media library image style.
+ */
+function oe_showcase_post_update_00022(): void {
+  ConfigImporter::importSingle('profile', 'oe_showcase', '/config/post_updates/00023_media_library', 'image.style.media_entity_browser_thumbnail');
+}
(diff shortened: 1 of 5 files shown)


brummbar (dismissed):
I would like to have a few improvements:

- remove a bit of borders, they are a bit everywhere:
image
Let's keep them only around.
- I pushed an update to the image style so the ratio is kept in all tabs. I'd like it looks like more like this
image
- let's also add this interaction
https://user-images.githubusercontent.com/8488617/201397222-20077784-bf4d-4062-9075-41fadeabd04a.mov
You can find some of these styles here build/core/themes/seven/css/theme/media-library.css
If you need help to have a live demo tell me.

tibi2303:
@brummbar 
image


I pushed now the changes, on my side it look like that, the left one has the hover effect.

oe_showcase.post_update.php
```diff
@@ -535,3 +535,10 @@ function oe_showcase_post_update_00021(): void {
   $editor->grantPermission('access images_and_videos entity browser pages');
   $editor->save();
 }
+
+/**
+ * Update the media library image style.
+ */
+function oe_showcase_post_update_00022(): void {
+  ConfigImporter::importSingle('profile', 'oe_showcase', '/config/post_updates/00023_media_library', 'image.style.media_entity_browser_thumbnail');
```
drishu, line 543: I know why this time it's `00023_media_library` but I think we should name it `00022_media_library` (depending on the publication, we might actually change the function name to 00023)

drishu (approved):

PR #202, opened 2022-11-25
OEL-1807: Install the publication content type.

brummbar:


composer.json
```diff
@@ -23,6 +23,7 @@
         "drupal/extra_field": "^2.1",
         "drupal/facets": "^2.0.5",
         "drupal/facets_form": "1.0.0-alpha3",
+        "drupal/field_display_label": "^1.3",
```
drishu: as per feedback from BA we should drop this

oe_showcase.profile
```diff
@@ -39,6 +41,34 @@ function oe_showcase_disable_taxonomy_term_view(array &$install_state): void {
...
+ * In order to import the fields, the node bundle needs to be imported, but
+ * if oe_starter_content_publication is installed, Drupal will try to import the
+ * overridden form/view modes which depend on the new field.
+ * To solve this, we need to move these files out of the install/optional config
+ * folders and have them installed "manually" in this install task.
+ *
+ * @param array $install_state
+ *   An array of information about the current installation state.
+ */
+function oe_showcase_import_overrides(array &$install_state): void {
```
drishu, line 63: this is ok in a profile, but is it ok in a subsite developer context?

tests/src/Traits/WysiwygTrait.php
```diff
@@ -0,0 +1,128 @@
...
+  /**
+   * Checks whether a WYSIWYG editor with the given field label is present.
+   *
+   * @param string $field
+   *   The label of the field to which the WYSIWYG editor is attached.
+   *
+   * @return bool
+   *   TRUE if the editor is present, FALSE otherwise.
+   */
+  protected function hasWysiwyg($field) {
```
drishu: I get that it may be useful in the future, but I would not add it just for this reason

tests/src/ExistingSiteJavascript/PublicationTest.php
```diff
@@ -0,0 +1,190 @@
...
+
+    $this->getMediaBrowserTileByMediaName('Euro with miniature figurines');
+    $this->getMediaBrowserTileByMediaName('Image title')->click();
+    $assert_session->buttonExists('Select media')->press();
+    $this->getSession()->switchToIFrame();
+    $assert_session->assertWaitOnAjaxRequest();
+    $assert_session->pageTextContains('Image title');
+
+    $assert_session->fieldExists('Reference code')->setValue('I-9876987012');
+    // @todo Assert formats.
```
drishu, line 93: I'll give it a try, but if not, I'll create a follow up with low prio

drishu (changes requested):

tests/src/ExistingSiteJavascript/PublicationTest.php
drishu: Date test implemented, the format we solve with the foramt correction PR after we merge this

drishu (approved):

PR #210, opened 2023-01-13
OEL-1574: Enable slim_select.

drishu:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/modules/oe_showcase_list_pages/tests/src/ExistingSiteJavascript/ListPagesTest.php
+++ b/modules/oe_showcase_list_pages/tests/src/ExistingSiteJavascript/ListPagesTest.php
@@ -418,7 +418,7 @@ class ListPagesTest extends ShowcaseExistingSiteJavascriptTestBase {
     // Filter results by location.
     $filter_form->pressButton('Clear filters');
     $location = $filter_form->findField('Location');
-    $location->selectOption('France');
+    $this->selectSlimOption($location, 'France');
     $this->scrollIntoView('#edit-submit');
     $search_button->click();
     $this->assertResultsTitle('Event List Page', 2);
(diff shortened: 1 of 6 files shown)


modules/oe_showcase_list_pages/tests/src/ExistingSiteJavascript/ListPagesTest.php
```diff
@@ -968,7 +960,32 @@ protected function createListPage(string $title, string $bundle, array $exposed_
...
+   *
+   * @param \Behat\Mink\Element\NodeElement $field
+   *   The form field.
+   * @param string $option
+   *   The option to select.
+   * @param bool $multiple
+   *   If old values should be kept.
+   */
+  protected function selectSlimOption(NodeElement $field, string $option, bool $multiple = FALSE): void {
+    $slim_select = $field->getParent()->find('css', 'div.ss-main');
```
brummbar: I think this should fit in a trait?

modules/oe_showcase_list_pages/tests/src/ExistingSiteJavascript/ListPagesTest.php
```diff
@@ -968,7 +960,32 @@ protected function createListPage(string $title, string $bundle, array $exposed_
...
+    if (!$multiple) {
+      $items = $slim_select->findAll('css', '.ss-value-delete');
+      foreach ($items as $item) {
+        $item->click();
+      }
+    }
+    $slim_select->click();
+    $this->getSession()->wait(1000);
+    $slim_select->find('xpath', '//div[contains(@class, "ss-option") and text()="' . $option . '"]')->click();
+    $this->getSession()->wait(1000);
```
brummbar: Uhm we should wait until an element is visible, like `.ss-list`.
Then we can wait until the `.ss-values` element contains an entry with the option selected.

brummbar (changes requested):

modules/oe_showcase_list_pages/tests/src/ExistingSiteJavascript/ListPagesTest.php
drishu: you'r right sorry, got lazy and copy pasted

brummbar (approved):

PR #211, opened 2023-01-17
OEL-2162: Project updates

drishu:
### Description

Enabled share block on project pages, removed total budget till we have a number widget.

### Change log

- Changed: Share block config
- Removed: Budget facet


modules/oe_showcase_list_pages/tests/src/ExistingSiteJavascript/ListPagesTest.php
```diff
@@ -541,17 +540,15 @@ public function testCreateListPages() {
       'Project pending',
     ]);
 
-    // Assert that the filter form for Events exists.
+    // Assert that the filter form for Projects exists.
     $filter_form = $assert_session->elementExists('css', '#oe-list-pages-facets-form');
     $filter_status = $filter_form->findField('Status');
     $filter_type = $filter_form->findField('Project type');
-    $filter_budget = $filter_form->findField('Total budget');
```
brummbar, line 548: Should we actually add a line to test that the budget is not there? Now the update path could be wrong and we still don't have the test failing :)

brummbar (changes requested):

brummbar (approved):

PR #212, opened 2023-01-17
OEL-2163: Content news required.

tibi2303:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/oe_showcase.post_update.php
+++ b/oe_showcase.post_update.php
@@ -704,11 +704,3 @@ function oe_showcase_post_update_00026(): void {
 function oe_showcase_post_update_00027(): void {
   \Drupal::service('module_installer')->install(['slim_select']);
 }
-
-/**
- * News content required.
- */
-function oe_showcase_post_update_00028(): void {
-  $field_config = FieldConfig::load('node.oe_sc_news.body');
-  $field_config->setRequired(TRUE)->save();
-}
--- a/oe_showcase.post_update.php
+++ b/oe_showcase.post_update.php
@@ -704,3 +704,11 @@ function oe_showcase_post_update_00026(): void {
 function oe_showcase_post_update_00027(): void {
   \Drupal::service('module_installer')->install(['slim_select']);
 }
+
+/**
+ * News content required.
+ */
+function oe_showcase_post_update_00028(): void {
+  $field_config = FieldConfig::load('node.oe_sc_news.body');
+  $field_config->setRequired(TRUE)->save();
+}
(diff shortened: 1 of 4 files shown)


oe_showcase.post_update.php
```diff
@@ -679,3 +679,11 @@ function oe_showcase_post_update_00024(): void {
 function oe_showcase_post_update_00025(): void {
   ConfigImporter::importSingle('profile', 'oe_showcase', '/config/post_updates/00025_event_location', 'core.entity_view_display.node.oe_sc_event.teaser');
 }
+
+/**
+ * News content required.
+ */
+function oe_showcase_post_update_00026(): void {
```
drishu: This is not needed here as we have the oe_starter_content one, we need to rather patch oe_starter_content in composer with a diff, so the new code gets used.

drishu (changes requested):

oe_showcase.post_update.php
tibi2303: done.

tests/src/ExistingSite/NewsTest.php
```diff
@@ -71,6 +71,12 @@ public function testCreateNews() {
...
     $this->drupalLogin($user);
 
+    // Assert that the content field is required.
+    $this->drupalGet('node/add/oe_sc_news');
+    $page->fillField('Title', 'Example title');
+    $page->pressButton('Save');
+    $assert_session->pageTextContains('Content field is required.');
+
     // Assert that editors have access to the Simple/Rich text formats.
     $this->drupalGet('node/add/oe_sc_news');
```
drishu: I'm 90% sure this line is not needed, drupal should stay on the form when it has errors

drishu (changes requested):

drishu (dismissed):

tests/src/ExistingSite/NewsTest.php
tibi2303: done.

drishu (dismissed):

oe_showcase.post_update.php
```diff
@@ -697,3 +697,11 @@ function oe_showcase_post_update_00026(): void {
     $block->save();
   }
 }
+
+/**
+ * News content required.
+ */
+function oe_showcase_post_update_00027(): void {
```
brummbar: Is this needed? We indeed have the one in starter content.

brummbar (changes requested):

oe_showcase.post_update.php
tibi2303: removed.

brummbar (approved):

PR #213, opened 2023-02-03
OEL-2050: Fixed user banner display.

tibi2303:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/themes/oe_showcase_theme/templates/user/user--full.html.twig
+++ b/themes/oe_showcase_theme/templates/user/user--full.html.twig
@@ -7,7 +7,6 @@
 {% apply spaceless %}
 <article{{ attributes }}>
   {# Combine margin and padding for precise spacing. #}
-  {# @todo Remove the 'mx-n3', 'px-1' once OEL-1226 is fixed. #}
   <div class="mx-md-4 w-100">
     {% set _field_bio %}
       <div class="pe-md-3-5">{{ content.field_bio|field_value }}</div>
@@ -19,7 +18,7 @@
         content.group_professional_information.field_organization|field_value,
         content.group_professional_information.field_current_position|field_value,
       ],
-      content: _field_bio|raw,
+      content: _field_bio,
       attributes: create_attribute().addClass('mb-4', 'pb-3'),
       card: {
         attributes: create_attribute().addClass("px-4")
--- a/themes/oe_showcase_theme/templates/user/user--full.html.twig
+++ b/themes/oe_showcase_theme/templates/user/user--full.html.twig
@@ -8,18 +8,26 @@
 <article{{ attributes }}>
   {# Combine margin and padding for precise spacing. #}
   {# @todo Remove the 'mx-n3', 'px-1' once OEL-1226 is fixed. #}
-  {{ pattern('content_banner', {
-    image: image,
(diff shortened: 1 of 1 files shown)


themes/oe_showcase_theme/templates/user/user--full.html.twig
```diff
@@ -8,18 +8,26 @@
 <article{{ attributes }}>
   {# Combine margin and padding for precise spacing. #}
   {# @todo Remove the 'mx-n3', 'px-1' once OEL-1226 is fixed. #}
```
drishu: since you removed the classes you can remove also the comment

themes/oe_showcase_theme/templates/user/user--full.html.twig
```diff
@@ -8,18 +8,26 @@
...
+      <div class="pe-md-3-5">{{ content.field_bio|field_value }}</div>
+    {% endset %}
+    {{ pattern('content_banner', {
+      image: image,
+      title: content.field_first_name|field_value|render ~ ' ' ~ content.field_last_name|field_value|render,
+      meta: [
+        content.group_professional_information.field_organization|field_value,
+        content.group_professional_information.field_current_position|field_value,
+      ],
+      content: _field_bio|raw,
```
drishu: the raw is not needed here

themes/oe_showcase_theme/templates/user/user--full.html.twig
```diff
@@ -8,18 +8,26 @@
...
-      content.group_professional_information.field_current_position|field_value,
-    ],
-    content: content.field_bio|field_value,
-    attributes: create_attribute().addClass('mb-4', 'pb-3', 'mx-n3', 'px-1'),
-  }) }}
-  {{ _self.description_list_section('Personal information', content.group_personal_information ?? {}) }}
-  {{ _self.description_list_section('Professional information', content.group_professional_information ?? {}) }}
+  <div class="mx-md-4 w-100">
+    {% set _field_bio %}
+      <div class="pe-md-3-5">{{ content.field_bio|field_value }}</div>
```
drishu, line 12: please confirm if you see any difference visually with or without this wrapper

drishu (changes requested):

themes/oe_showcase_theme/templates/user/user--full.html.twig
tibi2303: image
without the wrapper it will overflow the container, this limits the text width

themes/oe_showcase_theme/templates/user/user--full.html.twig
tibi2303: removed

themes/oe_showcase_theme/templates/user/user--full.html.twig
tibi2303: removed.

themes/oe_showcase_theme/templates/user/user--full.html.twig
drishu: So the idea is that without it the text might overflow, and the proper solution is to move it into the content banner block.

drishu (dismissed):

drishu (approved):

PR #219, opened 2023-06-16
OEL-2371: D10 compatibility and PHP 8.1 update.

drishu:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


modules/oe_showcase_search/config/install/search_api.index.showcase_search_index.yml
```diff
@@ -84,7 +84,7 @@ processor_settings:
       - body
       - rendered_item
       - title
-    ignorable: '[''¿¡!?,.:;]'
+    ignorable: "['¿¡!?,.:;]"
```
sergepavle, line 87: Seems like this change is missed in post_updates.

sergepavle (approved):

PR #235, opened 2024-01-10
OEL-2755: Disable the /node page.

AaronGilMartinez:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


oe_showcase.post_update.php
```diff
@@ -775,3 +775,15 @@ function oe_showcase_post_update_00032(): void {
   // Index elements.
   Index::load('oe_list_pages_index')->indexItems();
 }
+
+/**
+ * Disable frontpage view.
+ */
+function oe_showcase_post_update_00033(): void {
+  ConfigImporter::importSingle(
```
brummbar: Since we are just disabling the view, we can simply load it and disable it.

tests/src/ExistingSite/ProjectTest.php
```diff
@@ -197,12 +197,8 @@ public function testCreateProject() {
     $assert_session->pageTextContains('€22,90');
     $assert_session->pageTextContains('Project details');
 
-    // Check meta is present.
-    $this->drupalGet('/node');
-    $assert_session->elementExists('css', '.views-row:nth-child(1) .text-muted');
```
brummbar, line 202: Did you find what this test was supposed to cover?

brummbar (changes requested):

tests/src/ExistingSite/ProjectTest.php
AaronGilMartinez: I think it was there to test some information displayed in the teaser, or the teaser itself. Checking the file historic, it seems that in the past the test checked how project looked in lists. This test used node view or created a specific view only for tests.

While doing tests for other tasks I saw that sometimes modules add a test module/theme to support testing, but haven't come to a conclusion to test displays or templates, rather than testing the implementation of the visual element on the site or the support modules.

tests/src/ExistingSite/ProjectTest.php
```diff
@@ -243,10 +253,6 @@ public function testCreateProject() {
     $this->assertEquals('Participants', $participants);
 
     $this->assertSocialShareBlock();
-
-    $this->drupalGet('/node');
-    // Check meta is not present.
-    $assert_session->elementNotExists('css', '.views-row:nth-child(1) .text-muted');
```
drishu, line 249: think its fine to remove, stuff like this needs to be handled differently

drishu (dismissed):

drishu (approved):

brummbar (approved):

PR #237, opened 2024-01-30
OEL-2794: Update components to run on Drupal 10.2.x 

AaronGilMartinez:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -30,13 +30,12 @@ services:
       - SPARQL_UPDATE=true
       - DBA_PASSWORD=dba
   selenium:
-    image: registry.fpfis.eu/fpfis/selenium:standalone-chrome-3.141.59-oxygen
+    image: registry.fpfis.eu/fpfis/selenium:standalone-chrome-4.1.3-20220405
     shm_size: 2g
     environment:
       - DISPLAY=:99
-      - SE_OPTS=-debug
-      - SCREEN_WIDTH=1280
-      - SCREEN_HEIGHT=800
+      - SCREEN_WIDTH=1440
+      - SCREEN_HEIGHT=900
       - NODE_MAX_INSTANCES=5
       - NODE_MAX_SESSION=5
 
@@ -114,7 +113,7 @@ pipeline:
     image: registry.fpfis.eu/fpfis/httpd-php:8.1-ci
     commands:
       - ./vendor/bin/run ci:site-setup
-      - ./vendor/bin/drush en -y media_avportal_mock oe_media_oembed_mock js_testing_ajax_request_test js_testing_log_test
+      - ./vendor/bin/drush en -y media_avportal_mock oe_media_oembed_mock js_testing_ajax_request_test
       - ./vendor/bin/phpunit
 
 matrix:
--- a/.drone.yml
+++ b/.drone.yml
@@ -114,7 +114,7 @@ pipeline:
     image: registry.fpfis.eu/fpfis/httpd-php:8.1-ci
     commands:
       - ./vendor/bin/run ci:site-setup
-      - ./vendor/bin/drush en -y media_avportal_mock oe_media_oembed_mock
+      - ./vendor/bin/drush en -y media_avportal_mock oe_media_oembed_mock js_testing_ajax_request_test
       - ./vendor/bin/phpunit
 
 matrix:
--- a/.drone.yml
+++ b/.drone.yml
@@ -18,7 +18,7 @@ services:
     environment:
       - DOCUMENT_ROOT=/test/oe_showcase
   mysql:
-    image: registry.fpfis.eu/fpfis/sql:percona-5.7
+    image: git.fpfis.tech.ec.europa.eu/fpfis/dependency_proxy/containers/percona:5.7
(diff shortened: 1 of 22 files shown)


composer.json
```diff
@@ -14,7 +14,7 @@
         "drupal/block_field": "^1.0@RC",
         "drupal/ckeditor": "^1.0",
         "drupal/composite_reference": "^2.2",
-        "drupal/core-recommended": "^10.1",
+        "drupal/core-recommended": "^10.2",
```
drishu: try with ^10 and run composer update for these packages with the dependencies flag, it should install the latest 10.2

also I think the php requirement should be a minimum of php 8.1

you can remove oe_starter_content and oe_content_extra from the "repositories" section while we are here, they where added to packagist

tests/src/ExistingSiteJavascript/ShowcaseExistingSiteJavascriptTestBase.php
```diff
@@ -48,6 +48,7 @@ public function tearDown(): void {
     $this->restoreConfiguration();
 
     parent::tearDown();
+
```
drishu: remove

.drone.yml
```diff
@@ -114,7 +112,7 @@ pipeline:
     image: registry.fpfis.eu/fpfis/httpd-php:8.1-ci
     commands:
       - ./vendor/bin/run ci:site-setup
-      - ./vendor/bin/drush en -y media_avportal_mock oe_media_oembed_mock
+      - ./vendor/bin/drush en -y media_avportal_mock oe_media_oembed_mock js_testing_ajax_request_test
```
drishu, line 115: intresting but I dont see the usage, can you explain to me a bit?

modules/oe_showcase_list_pages/tests/src/ExistingSiteJavascript/ListPagesTest.php
```diff
@@ -948,4 +937,15 @@ protected function scrollIntoView(string $selector): void {
     $this->getSession()->executeScript("window.scrollBy({top: document.querySelector('$selector').offsetTop, left: 0, behavior: 'instant'})");
   }
 
+  /**
+   * Select a sort option and waits until is applied.
+   *
+   * @param string $option
+   *   The option to be applied.
+   */
+  protected function sortByOption(string $option): void {
```
drishu: the name of the function does not reflect what it does, this function selects an option, so selectSortByOption

drishu (changes requested):
We can also safely drop version 9 from all core_version_requirement

modules/oe_showcase_list_pages/tests/src/ExistingSiteJavascript/ListPagesTest.php
AaronGilMartinez: Renamed

tests/src/ExistingSiteJavascript/ShowcaseExistingSiteJavascriptTestBase.php
AaronGilMartinez: done

composer.json
AaronGilMartinez: done

.drone.yml
AaronGilMartinez: The core Javascript tests base class (core/tests/Drupal/FunctionalJavascriptTests/WebDriverTestBase.php) uses that module to track Ajax requests, with Drupal 10.2 there are changes and new checks on the function assertWaitOnAjaxRequest (drupal-site/web/core/tests/Drupal/FunctionalJavascriptTests/JSWebAssert.php) and if there is no Ajax request the test fails.
Our Javascript tests extends the provided by weitzman/drupal-test-site so the module is not installed and some test fail.

drishu (dismissed):

drishu (approved):

composer.json
```diff
@@ -112,7 +105,7 @@
                 "Allow altering the post_update run order @see https://www.drupal.org/project/drupal/issues/3129231": "https://www.drupal.org/files/issues/2022-06-27/drupal-post_update_alter-3129231-20.patch",
                 "Unserialize(): Passing null to parameter #1 ($data) of type string is deprecated @see https://www.drupal.org/project/drupal/issues/3300404": "https://www.drupal.org/files/issues/2023-03-03/3300404-28-map-field-failing-php-81-combined.patch",
                 "strpos(): Passing null is deprecated in core/lib/Drupal/Core/Mail/Plugin/Mail/PhpMail.php @see https://www.drupal.org/project/drupal/issues/3328187": "https://www.drupal.org/files/issues/2023-07-05/3372667-2_0.patch",
-                "Simple decimals fail to pass validation @see https://www.drupal.org/project/drupal/issues/2230909": "https://www.drupal.org/files/issues/2022-11-19/2230909-285-D10.patch"
+                "Simple decimals fail to pass validation @see https://www.drupal.org/project/drupal/issues/2230909": "https://www.drupal.org/files/issues/2023-12-21/2230909-309.patch"
```
drishu: think we should use the stable oe_content version and drop this line

drishu (dismissed):

composer.json
AaronGilMartinez: Removed and updated

brummbar (approved):

PR #238, opened 2024-02-14
OEL-2592: Changed form filter submit label.

tibi2303:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/oe_showcase.profile
+++ b/oe_showcase.profile
@@ -237,3 +237,11 @@ function oe_showcase_element_info_alter(array &$types) {
     $types['text_format']['#process'][] = [AllowedFormats::class, 'alterTextFormatHelp'];
   }
 }
+
+/**
+ * Implements hook_form_FORM_ID_alter().
+ */
+function oe_showcase_form_oe_list_pages_facets_form_alter(&$form, &$form_state, $form_id) {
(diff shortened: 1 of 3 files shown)


oe_showcase.profile
```diff
@@ -231,3 +231,11 @@ function oe_showcase_element_info_alter(array &$types) {
     $types['text_format']['#process'][] = [AllowedFormats::class, 'alterTextFormatHelp'];
   }
 }
+
+/**
+ * Implements hook_form_FORM_ID_alter().
+ */
+function oe_showcase_form_oe_list_pages_facets_form_alter(&$form, &$form_state, $form_id) {
```
AaronGilMartinez: I think could be a good idea to set this in the oe_showcase_list_pages module

AaronGilMartinez (changes requested):

oe_showcase.profile
AaronGilMartinez: Drone is failing, test needs to be updated, and branch too. Can you check, please?

AaronGilMartinez (approved):

PR #239, opened 2024-02-16
OEL-2617: Changed label from Title to Name in person list page.

tibi2303:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


config/post_updates/00016_event_type_and_filters/facets.facet.oelp_oe_sc_person__title.yml
```diff
@@ -2,7 +2,7 @@ langcode: en
 status: true
 dependencies: {  }
 id: oelp_oe_sc_person__title
-name: Title
+name: Name
```
drishu: Don't change this, this was already executed and will not be executed again

modules/oe_showcase_list_pages/config/install/facets.facet.oelp_oe_sc_person__title.yml
```diff
@@ -1,8 +1,10 @@
 langcode: en
 status: true
 dependencies: {  }
+_core:
```
drishu: please change the title in the admin (UI) and use drush cde oe_showcase_list_pages to export it, the copy the contents in the update

drishu (changes requested):
The tests are failing and this is good, because we do cover this facet in tests so you need to update that too

drishu (approved):

PR #240, opened 2024-02-23
OEL-2616: The selection of gender is insulting for users who could not be identifying as CIS female/male.

AaronGilMartinez:
…t be identifying as CIS female/male.

### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


modules/oe_showcase_user_profile/oe_showcase_user_profile.install
```diff
@@ -18,7 +18,6 @@ function oe_showcase_user_profile_install($is_syncing): void {
   $graphs = [
     'country' => 'http://publications.europa.eu/resource/authority/country',
     'language' => 'http://publications.europa.eu/resource/authority/language',
-    'human-sex' => 'http://publications.europa.eu/resource/dataset/human-sex',
```
AaronGilMartinez, line 21: @todo Check if the hook update has to remove this

AaronGilMartinez: Although the graph is used by oe_person from oe_content, the field is not present in branch 1.x.

brummbar: It's ok to leave the graph.

modules/oe_showcase_user_profile/tests/src/ExistingSite/IntegrationTest.php
```diff
@@ -34,11 +34,17 @@ public function testIntegration(): void {
     ]);
     $this->markEntityForCleanup($user);
 
+    // Check that the same number of defined fields are present in the form.
     $this->drupalGet("user/{$user->id()}/edit");
+    $fields = array_filter($user->getFieldDefinitions(), function ($field) {
```
brummbar: Why this? It's enough to test that we don't have a gender field anymore.

modules/oe_showcase_user_profile/oe_showcase_user_profile.post_update.php
```diff
@@ -0,0 +1,50 @@
...
+use Drupal\oe_bootstrap_theme\ConfigImporter;
+
+/**
+ * Remove gender field.
+ */
+function oe_showcase_user_profile_post_update_00001(&$sandbox): void {
+  FieldConfig::loadByName('user', 'user', 'field_gender')->delete();
+
+  // Remove rdf skos graph.
+  $graph = [
```
brummbar: We can leave the graph, it's fine.

brummbar (changes requested):

modules/oe_showcase_user_profile/tests/src/ExistingSite/IntegrationTest.php
AaronGilMartinez: Wanted to avoid checking something that is not there, and try to check the total fields that should be there. The field definition variable its to make sure it has been deleted not hidden.

modules/oe_showcase_user_profile/oe_showcase_user_profile.post_update.php
AaronGilMartinez: Ok

modules/oe_showcase_user_profile/tests/src/ExistingSite/IntegrationTest.php
brummbar: It's ok if it would be hidden. If you really want to check if the entity bundle doesn't have the field, you can add a one liner at the end to check programmatically that field X does not exists.

brummbar (changes requested):

modules/oe_showcase_user_profile/tests/src/ExistingSite/IntegrationTest.php
AaronGilMartinez: Added the field check

brummbar (approved):

PR #248, opened 2024-04-25
OEL-2929: Add an optional title, description and image on the listing page.

AaronGilMartinez:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/config/install/core.entity_form_display.node.oe_list_page.default.yml
+++ b/config/install/core.entity_form_display.node.oe_list_page.default.yml
@@ -2,9 +2,13 @@ langcode: en
 status: true
 dependencies:
   config:
+    - field.field.node.oe_list_page.oe_featured_media
+    - field.field.node.oe_list_page.oe_summary
     - node.type.oe_list_page
   module:
+    - oe_content_featured_media_field
     - path
+    - text
 id: node.oe_list_page.default
 targetEntityType: node
 bundle: oe_list_page
@@ -12,7 +16,7 @@ mode: default
 content:
   created:
     type: datetime_timestamp
-    weight: 3
+    weight: 5
     region: content
     settings: {  }
     third_party_settings: {  }
@@ -23,29 +27,47 @@ content:
     settings:
       include_locked: true
     third_party_settings: {  }
+  oe_featured_media:
+    type: oe_featured_media_autocomplete
+    weight: 2
+    region: content
+    settings:
+      match_operator: CONTAINS
+      match_limit: 10
+      size: 60
+      placeholder: ''
+    third_party_settings: {  }
+  oe_summary:
+    type: text_textarea
+    weight: 3
+    region: content
+    settings:
+      rows: 5
+      placeholder: ''
+    third_party_settings: {  }
   path:
     type: path
-    weight: 30
+    weight: 9
     region: content
     settings: {  }
     third_party_settings: {  }
   promote:
-    type: boolean_checkbox
-    weight: 4
-    region: content
-    settings:
-      display_label: true
-    third_party_settings: {  }
-  status:
     type: boolean_checkbox
     weight: 6
     region: content
     settings:
       display_label: true
     third_party_settings: {  }
+  status:
+    type: boolean_checkbox
+    weight: 8
+    region: content
+    settings:
+      display_label: true
+    third_party_settings: {  }
   sticky:
     type: boolean_checkbox
-    weight: 5
+    weight: 7
     region: content
     settings:
       display_label: true
@@ -60,7 +82,7 @@ content:
     third_party_settings: {  }
   uid:
     type: entity_reference_autocomplete
-    weight: 2
+    weight: 4
     region: content
     settings:
       match_operator: CONTAINS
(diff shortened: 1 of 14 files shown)


config/install/core.entity_form_display.node.oe_list_page.default.yml
```diff
@@ -23,29 +27,47 @@ content:
     settings:
       include_locked: true
     third_party_settings: {  }
+  oe_featured_media:
+    type: oe_featured_media_autocomplete
```
brummbar: We can use the other widget here, so it's easier to pick images.

config/install/field.field.node.oe_list_page.oe_summary.yml
```diff
@@ -0,0 +1,23 @@
...
+bundle: oe_list_page
+label: Summary
+description: 'A short text that will be displayed below the page title. This should be a brief summary of the content on the page that tells the user what information they will find on this page.'
+required: false
+translatable: false
+default_value: {  }
+default_value_callback: ''
+settings:
+  allowed_formats:
+    - rich_text
```
brummbar: This should be simple rich text.

modules/oe_showcase_list_pages/tests/src/ExistingSiteJavascript/ListPagesTest.php
```diff
@@ -299,7 +316,7 @@ public function testCreateListPages() {
     ]);
 
     $this->selectSortByOption('A-Z');
-    $this->assertResultsTitle('News List Page', 12);
+    $this->assertResultsTitle(12);
```
brummbar: It's not a title anymore, it's a count.

modules/oe_showcase_list_pages/tests/src/ExistingSiteJavascript/ListPagesTest.php
```diff
@@ -932,32 +966,67 @@ protected function assertElementsTexts(array $expected, array $elements): void {
...
+    $media_name = '';
 
+    if ($has_media) {
+      $file = File::create([
+        'uri' => $this->getTestFiles('image')[0]->uri,
+      ]);
+      $file->save();
+
+      // Create a sample image media entity to be embedded.
+      $media = Media::create([
```
brummbar: We have a media creation trait.

modules/oe_showcase_list_pages/tests/src/ExistingSiteJavascript/ListPagesTest.php
```diff
@@ -974,4 +1043,41 @@ protected function selectSortByOption(string $option): void {
...
+   * Asserts content banner in the page.
+   *
+   * @param string $expected_title
+   *   The expected title text.
+   * @param string $expected_summary
+   *   The expected summary text.
+   * @param array $expected_image
+   *   Associative array with expected image attributes.
+   */
+  protected function assertPageBanner(string $expected_title, string $expected_summary = '', array $expected_image = []) {
```
brummbar: Why not ContentBannerAssert?

brummbar (changes requested):

modules/oe_showcase_list_pages/tests/src/ExistingSiteJavascript/ListPagesTest.php
AaronGilMartinez: The function name can be more descriptive indeed, we should use lowerCamelCase and assert functions should start with 'assert' prefix like: assertContentBanner

modules/oe_showcase_list_pages/tests/src/ExistingSiteJavascript/ListPagesTest.php
AaronGilMartinez: We can reuse it, we have some other content types tests where we create media instead of using the trait I think would be a good idea to apply to all of them

modules/oe_showcase_list_pages/tests/src/ExistingSiteJavascript/ListPagesTest.php
AaronGilMartinez: True

modules/oe_showcase_list_pages/tests/src/ExistingSiteJavascript/ListPagesTest.php
AaronGilMartinez: Adding it only for list pages until confirmation

config/install/field.field.node.oe_list_page.oe_summary.yml
AaronGilMartinez: Changed

modules/oe_showcase_list_pages/tests/src/ExistingSiteJavascript/ListPagesTest.php
AaronGilMartinez: Missunderstood, replacing with ContentBannerAssert and assertPattern()

config/install/core.entity_form_display.node.oe_list_page.default.yml
AaronGilMartinez: Added entity browser like in publications

brummbar (approved):

PR #252, opened 2024-05-28
OEL-2604: Change alignment of the sorting options.

AaronGilMartinez:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


drishu (changes requested):
it solves the issue but I'm not convinced it brings any value, pleople would look at OESA think it's solved and then when they install a project they find out they need to fix it on project level.

I'd rather do something like https://github.com/openeuropa/oe_bootstrap_theme/blob/1.x/theme-settings.php and then on 2.x remove them

brummbar:
Not to change in here, it's to be fixed in oe_whitelabel.

PR #255, opened 2024-06-06
OEL-2606: Add hook update to enable glossary terms automatic alias.

AaronGilMartinez:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/modules/oe_showcase_glossary/oe_showcase_glossary.post_update.php
+++ b/modules/oe_showcase_glossary/oe_showcase_glossary.post_update.php
@@ -8,10 +8,46 @@
 declare(strict_types=1);
 
 use Drupal\oe_bootstrap_theme\ConfigImporter;
+use Drupal\taxonomy\Entity\Term;
 
 /**
- * Add alias for Glossary term.
+ * Add alias for Glossary term and enable automatic alias for existing terms.
  */
-function oe_showcase_glossary_post_update_00001(): void {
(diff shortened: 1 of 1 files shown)


modules/oe_showcase_glossary/oe_showcase_glossary.post_update.php
```diff
@@ -8,10 +8,46 @@
...
+  $tids = \Drupal::entityQuery('taxonomy_term')
+    ->range($sandbox['current'], 50)
+    ->condition('vid', 'glossary')
+    ->accessCheck(FALSE)
+    ->execute();
+
+  foreach ($tids as $tid) {
+    $term = Term::load($tid);
+    // Enable automatic alias generation.
+    $term->path->pathauto = TRUE;
```
drishu: my editor indicates path is protected, lets try something like `$term->set('path', ['pathauto' => TRUE]);`

drishu (changes requested):

modules/oe_showcase_glossary/oe_showcase_glossary.post_update.php
AaronGilMartinez: Indeed, its better

drishu (approved):

intelektron (approved):

PR #261, opened 2024-07-16
OEL-3057: Update to Drupal 10.3

AaronGilMartinez:


composer.lock
```diff
@@ -128,16 +128,16 @@
         },
         {
             "name": "composer/semver",
-            "version": "3.4.0",
```
drishu, line 131: seems that a lot was updated, maybe unnecessary, I think you'll have better results with composer update drupal/core-recommended drupal/core-composer-scaffold drupal/core-dev -W

drishu (changes requested):

composer.lock
AaronGilMartinez: Changed, updated core, media_avportal and datetime_testing

drishu (approved):

PR #263, opened 2024-07-28
OEL-3124: Add url alias patterns to content types.

drishu:
### Description

Add url alias patterns to content types that dont have it.

### Change log

- Added: Url alias patterns
- Changed: tests, subscription email template in config
- Fixed: tests


modules/oe_showcase_subscriptions/tests/src/ExistingSite/EventSubscriptionTest.php
```diff
@@ -96,13 +96,13 @@ public function testEventNotifications(): void {
 
     $this->assertMail($authenticated_user->getEmail(), ['The event Event update 1 has been updated.'], [
       [
-        'url' => $event->toUrl()->setAbsolute()->toString(),
+        'url' => $event->toUrl()->setOption('alias', TRUE)->setAbsolute()->toString(),
```
drishu, line 99: setting alias to TRUE to get unaliased links is not confusing at all :)

oe_showcase.post_update.php
```diff
@@ -896,3 +896,16 @@ function oe_showcase_post_update_00040(): void {
...
+ * Add Pathauto patterns to content types.
+ */
+function oe_showcase_post_update_00041(): void {
+  ConfigImporter::importMultiple('profile', 'oe_showcase', '/config/post_updates/00041_pathauto_patterns', [
+    'message.template.node_event_update',
+    'pathauto.pattern.event_url_alias_pattern',
+    'pathauto.pattern.news_url_alias_pattern',
+    'pathauto.pattern.person_url_alias_pattern',
+    'pathauto.pattern.publication_url_alias_pattern',
+  ]);
```
AaronGilMartinez, line 932: Missing hook_update to regenerate content aliases like: https://github.com/openeuropa/oe_showcase/blob/18c2e6b68ec2284b6425d780a92e14338d22defe/modules/oe_showcase_glossary/oe_showcase_glossary.post_update.php#L13-L52

AaronGilMartinez: Current pattern alias for project and page doesn't match with requirements, both need update.
See table in https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-2596

AaronGilMartinez (changes requested):

modules/oe_showcase_subscriptions/tests/src/ExistingSite/EventSubscriptionTest.php
AaronGilMartinez: Another option would be `'path_processing' => FALSE`

drishu: I'm not certain about your comment, I checked core and found instances of how I did it, but not on how you suggested. I am aware of the option but dont know if all possible plugins should be shut down.

oe_showcase.post_update.php
drishu: good find, thx. I didn't add the pages to the update, as I dont think we should touch those old pages (News, Events, etc I think is ok, as they have listing pages and can be found like that).

AaronGilMartinez: About pages I would add them, why not?
List pages will need to be updated, pattern is there, but existing content isn't using alias pattern.

modules/oe_showcase_subscriptions/tests/src/ExistingSite/EventSubscriptionTest.php
AaronGilMartinez: To complement information, I found this option in \Drupal\Core\Routing\UrlGenerator
I though was interesting to mention

AaronGilMartinez (changes requested):

modules/oe_showcase_subscriptions/tests/src/ExistingSite/EventSubscriptionTest.php
drishu: I'm aware, but there is not usage like that in core, if we check /Drupal/Core/Entity/Controller/EntityViewController.php we see this option in use

AaronGilMartinez (dismissed):

AaronGilMartinez (changes requested):

oe_showcase.post_update.php
```diff
@@ -990,6 +990,7 @@ function oe_showcase_post_update_00042(array &$sandbox): void {
     $node = Node::load($node);
     // Enable automatic alias generation.
     $node->set('path', ['pathauto' => TRUE]);
+    $node->search_api_skip_tracking = TRUE;
```
AaronGilMartinez, line 993: I though was needed for home too, but its a change of the content and shouldn't be skipped, even with warning.

AaronGilMartinez (approved):

brummbar (approved):

PR #270, opened 2024-09-09
OEL-3168: Send subscriptions emails after update hooks run

AaronGilMartinez:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


modules/oe_showcase_subscriptions/oe_showcase_subscriptions.module
```diff
@@ -10,9 +10,9 @@ declare(strict_types=1);
 use Drupal\node\NodeInterface;
 
 /**
- * Implements hook_node_update().
+ * Implements hook_node_postsave().
  */
-function oe_showcase_subscriptions_node_update(NodeInterface $node) {
+function oe_showcase_subscriptions_node_postsave(NodeInterface $node) {
```
donquixote, line 26: I can see how this would work.
But we can achieve the same with hook_node_update() + hook_module_implements_alter() to make it run last(ish), without a patch.
The idea of a separate hook is not bad, but relying on the patch introduces fragility which we could avoid.

modules/oe_showcase_subscriptions/tests/src/ExistingSite/EventSubscriptionTest.php
```diff
@@ -121,13 +122,14 @@ public function testEventNotifications(): void {
...
     $assert_session->pageTextContains('Event Event update 2 has been updated.');
+    \Drupal::service('path_alias.manager')->cacheClear('node/' . $event->id());
     $this->drupalLogout();
 
     // Only one email is expected, to the anonymous user who didn't opt for a
     // digest.
     $this->assertMail('test_anon@example.com', ['The event Event update 2 has been updated.'], [
       [
-        'url' => $event->toUrl()->setOption('alias', TRUE)->setAbsolute()->toString(),
+        'url' => $event->toUrl()->setAbsolute()->toString(),
```
donquixote: Could we use an explicit literal string here?
The alias should be always 'event-update-2' at this point, based on the title update from earlier.
Using the explicit string makes the test more predictable.

donquixote: Not so sure about this anymore.
On the first test run and even on further repeated successful runs the url alias seems to be stable.
Do we need to consider scenarios where it is not?

modules/oe_showcase_subscriptions/oe_showcase_subscriptions.module
donquixote: I tried this, it's a dead end.
hook_entity_update() runs after hook_node_update(), and hook_module_implements_alter() cannot change that.
So we either use hook_node_postsave() with the patch, or we use hook_entity_update() and check type = node.

My conclusion is we should keep postsave.

modules/oe_showcase_subscriptions/tests/src/ExistingSite/EventSubscriptionTest.php
```diff
@@ -162,6 +164,7 @@ public function testEventNotifications(): void {
     $assert_session->fieldExists('Title')->setValue('Event update 3');
     $assert_session->buttonExists('Save')->press();
     $assert_session->pageTextContains('Event Event update 3 has been updated.');
+    \Drupal::service('path_alias.manager')->cacheClear('node/' . $event->id());
```
donquixote: The reason why this cache clear is needed:
The url in the email is correct.
But the url from `$event->toUrl()->setAbsolute()->toString()` is the old one from before saving.
The alias storage and persistent cache already has the correct alias, but the cache property `AliasManager::$lookupMap` still has the old value.

If we work with explicit literal aliases in the assertions, this cacheClear() call can be dropped - TBD.

modules/oe_showcase_subscriptions/tests/src/ExistingSite/EventSubscriptionTest.php
donquixote: I have two commits that bring you the literal strings and remove the clear cache.
I cannot push for some reason..
Ping me and we can look at it together.

modules/oe_showcase_subscriptions/tests/src/ExistingSite/EventSubscriptionTest.php
```diff
@@ -94,15 +95,20 @@ public function testEventNotifications(): void {
     $assert_session->buttonExists('Save')->press();
     $assert_session->pageTextContains('Event Event update 1 has been updated.');
 
+    // Construct urls with explicit path and fully transparent construction
+    // method (string concatenation), to be 100% sure that urls found in emails
+    // use the updated url alias.
+    $localized_base_url = Url::fromUserInput('/')->setAbsolute()->toString();
+    $make_url = fn (string $path): string => $localized_base_url . $path;
```
drishu: - making a function just for a concatanation is confusing,
- drone is red

modules/oe_showcase_subscriptions/oe_showcase_subscriptions.module
drishu: yes, unfortunately there is no hook after entity_update so we have to use this

drishu (changes requested):

modules/oe_showcase_subscriptions/tests/src/ExistingSite/EventSubscriptionTest.php
AaronGilMartinez: @donquixote did changes and drone green

drishu (approved):

PR #277, opened 2024-10-15
OEL-3424: Added h1 tags on search and homepage.

tibi2303:


themes/oe_showcase_theme/oe_showcase_theme.theme
```diff
@@ -82,3 +82,17 @@ function oe_showcase_theme_preprocess_user__full(array &$variables): void {
 function oe_showcase_theme_preprocess_media__document__showcase_embed(&$variables): void {
   oe_whitelabel_preprocess_media__document__default($variables);
 }
+
+/**
+ * Implements hook_preprocess_page().
+ *
+ * Adds an H1 heading to the homepage and search pages.
+ */
+function oe_showcase_theme_preprocess_page(array &$variables): void {
```
drishu, line 91: since the ticket asked for "We don't use the default block" then this is ok, only one problem, if I visit the front page I dont see any h1

modules/oe_showcase_search/tests/src/ExistingSiteJavascript/SearchFiltersTest.php
```diff
@@ -113,6 +113,7 @@ public function testSearchFilters() {
     $this->indexItems('showcase_search_index');
     $this->drupalGet('/search');
 
+    $search_title = $assert_session->elementExists('css', 'h1.visually-hidden');
```
drishu: $search_title is not used

themes/oe_showcase_theme/templates/node/node--search-result.html.twig
```diff
@@ -16,6 +16,7 @@
 {{ pattern('card', {
   variant: 'search',
   title: _title,
+  title_tag: 'h2',
```
drishu, line 19: composer lock has 1.16.0 for oebt but I still see h2 in the results list

tests/src/ExistingSite/FrontPageHeadingTest.php
```diff
@@ -0,0 +1,41 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function setUp(): void {
+    parent::setUp();
+
+    // Set /node/1 as the front page to not get redirected.
+    $config_factory = \Drupal::service('config.factory')->getEditable('system.site');
+    $config_factory->set('page.front', '/node/1')->save();
```
drishu: I tried without, works

tests/src/ExistingSite/FrontPageHeadingTest.php
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  /**
+   * Asserts H1 on the front page.
+   */
+  public function testFrontPageHeading(): void {
+    $this->drupalGet('<front>');
+    $this->assertFrontPageH1();
```
drishu: you dont need a separate method as there wont be any repetition

tests/src/ExistingSite/FrontPageHeadingTest.php
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  /**
+   * Asserts that an H1 with the class visually-hidden exists.
+   */
+  private function assertFrontPageH1(): void {
+    $this->assertSession()->elementExists('css', 'h1.visually-hidden');
+
+    $h1 = $this->getSession()->getPage()->find('css', 'h1.visually-hidden');
+    $this->assertNotNull($h1);
```
drishu: these lines do the same thing twice

tests/src/ExistingSite/FrontPageHeadingTest.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drupal\Tests\oe_showcase\ExistingSite;
+
+/**
+ * Class to test the presence of H1 on the front page.
+ */
+class FrontPageHeadingTest extends ShowcaseExistingSiteTestBase {
```
drishu:
```suggestion
class FrontPageTest extends ShowcaseExistingSiteTestBase {
```
lets name the class more generic, we might add more stuff in the future, the method name can stay the same

drishu (changes requested):

themes/oe_showcase_theme/oe_showcase_theme.theme
tibi2303: yeah, on the dev env the front is a redirect... is not actually front.. if you dump the var in this function you will see is_front -> false.

themes/oe_showcase_theme/templates/node/node--search-result.html.twig
tibi2303: that's for the main cards in the listing, not for the main heading, they were h1...

tests/src/ExistingSite/FrontPageHeadingTest.php
tibi2303: on drone it fails.. tested on a new env and it's failing idk..

tests/src/ExistingSite/FrontPageHeadingTest.php
tibi2303: done.

tests/src/ExistingSite/FrontPageHeadingTest.php
tibi2303: done.

tests/src/ExistingSite/FrontPageHeadingTest.php
tibi2303: done.

modules/oe_showcase_search/tests/src/ExistingSiteJavascript/SearchFiltersTest.php
tibi2303: done.

modules/oe_showcase_default_content/content/node/f63f9269-b08b-47dc-ad12-5fa53a026a54.yml
```diff
@@ -33,7 +33,7 @@ default:
     -
       alias: /home
       langcode: en
-      pathauto: 0
+      pathauto: 1
```
drishu: this is not needed

tests/src/ExistingSite/FrontPageTest.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drupal\Tests\oe_showcase\ExistingSite;
+
+/**
+ * Class to test the presence of H1 on the front page.
```
drishu: update also he class comment

themes/oe_showcase_theme/templates/node/node--search-result.html.twig
drishu: no, I mean the card titles are still h1

config/install/system.site.yml
```diff
@@ -1,2 +1,2 @@
 page:
-  front: /home
+  front: /node/5
```
drishu, line 2: normally when working in projects you would set system.site on prod and ignore this config in dev, but here we have a different situation, we need this file, but we can't hardcode node IDs. The issue is that the url alias here needs to be normalized, so lets add another task in oe_showcase_install_tasks_alter and normalize like its done in SiteInformationForm::validateForm(), like this, every install this will be normalized

drishu (changes requested):
the reason the update fails in drone is because you alter configuration for fresh install but dont provide an update path for existing installations

config/install/system.site.yml
drishu: seems oe_showcase_install_tasks_alter runs before config install

drishu (approved):

PR #279, opened 2024-11-06
Prepare release 1.13.0

tibi2303:


composer.lock
```diff
@@ -9864,16 +9864,16 @@
         },
         {
             "name": "symfony/http-foundation",
-            "version": "v6.4.12",
+            "version": "v6.4.14",
```
drishu, line 9867: nice, but we need to update the oe_bootstrap_theme oe_whitelabel -W

drishu (approved):

PR #283, opened 2025-02-18
OEL-3504: Update to slim_select 2

piotrsmykaj:


oe_showcase.post_update.php
```diff
@@ -1026,3 +1026,14 @@ function oe_showcase_post_update_00044(): void {
...
+
+/**
+ * Update Slim select to version 2.
+ */
+function oe_showcase_post_update_00045(): void {
+  $storage = new FileStorage(\Drupal::service('extension.list.profile')->getPath('oe_showcase') . '/config/post_updates/00045_slim_select');
+  $data = $storage->read('slim_select.settings');
+
+  $config_factory = \Drupal::configFactory();
+  $config_factory->getEditable('slim_select.settings')->setData($data)->save();
```
drishu, line 1038: for consistency please use the ConfigImporter class

also please add the slim select config permissions to an appropriate role (if no appropriate role is found then create a new one). Export the role and also update the role here

drishu (changes requested):

oe_showcase.post_update.php
piotrsmykaj: the ConfigImporter will not work as slim_select.settings is not an entity.

drishu (dismissed):

drishu (approved):

PR #304, opened 2025-05-23
OEL-3817: [oe_whitelabel] Preserve the query parameters when submitting the SearchForm

enriquelacoma:
PR created to test the change from whitelabel

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -62,7 +62,7 @@
         "openeuropa/oe_starter_content": "^1.5",
         "openeuropa/oe_subscriptions": "^1.0@beta",
         "openeuropa/oe_webtools": "^1.40",
-        "openeuropa/oe_whitelabel": "dev-OEL-3817"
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -59,7 +59,7 @@
         "openeuropa/oe_starter_content": "^1.2.0",
         "openeuropa/oe_subscriptions": "^1.0@alpha",
         "openeuropa/oe_webtools": "^1.23",
-        "openeuropa/oe_whitelabel": "^1.18"
+        "openeuropa/oe_whitelabel": "dev-OEL-3817"
```
tibi2303: we usually do the prerelease version like this: 0.{ticketnumber}.{timestamp} but leave it like this for now.

enriquelacoma: @tibi2303  I changed this to use a prereleaes a fix the phpcs error in drone

AaronGilMartinez (dismissed):

AaronGilMartinez (approved):

PR #305, opened 2025-05-27
OEL-3865: Update oe_oembed.

tibi2303:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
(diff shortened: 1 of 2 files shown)


composer.lock
```diff
@@ -2202,29 +2202,29 @@
         },
         {
             "name": "drupal/embed",
-            "version": "1.7.0",
+            "version": "1.10.0",
```
drishu, line 2205: this does not help, the latest version is 0.13, we should change this in oe_showcase and add the stream wrapper only if needed.

drishu (commented):
From my tests if the composer issues are resolved then the updb works as expected and the button in the editor will also be ok

PR #307, opened 2025-06-02
OEL-3865: Added remote_stream_wrapper module.

tibi2303:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,6 +34,7 @@
         "drupal/purge_users": "^3.3",
         "drupal/pwbi": "^2.0",
         "drupal/redirect": "^1.8",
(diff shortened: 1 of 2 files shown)


enriquelacoma:
@tibi2303 There is an error in the oesa pipeline https://git.fpfis.tech.ec.europa.eu/digit/digit-oesa-dev/-/jobs/8126214 because we are not using a prerelease "Package openeuropa/oe_showcase:dev-OEL-3865 cannot be used in dev version."

It worked on a local install

enriquelacoma (approved):

PR #311, opened 2025-06-13
OEL-3872: [oe_showcase] Update MySQL Percona 5.7 -> 8.0

enriquelacoma:


modules/oe_showcase_glossary/tests/src/ExistingSite/GlossaryTest.php
```diff
@@ -429,15 +429,18 @@ protected function generateTerms(): array {
       ]);
     }
 
-    // Sort the terms alphabetically.
+    // Sort the terms alphabetically.:
```
AaronGilMartinez: Is this colon ok?

AaronGilMartinez (changes requested):

modules/oe_showcase_glossary/tests/src/ExistingSite/GlossaryTest.php
enriquelacoma: nop :)

AaronGilMartinez (approved):

PR #312, opened 2025-06-14
OEL-3433: Move to ckeditor 5.

drishu:
### Description

Update config and tests to ckeditor 5.

Most of the config is just a drush cde oe_showcase export, you can see that just the order of some elements changed in the yaml config.

I disabled ckeditor 4 so that we can remove it in the release after this is released.


config/install/views.view.media_entity_browsers.yml
```diff
@@ -214,6 +214,7 @@ display:
             offset: false
             offset_label: Offset
           quantity: 9
+          pagination_heading_level: h4
```
donquixote, line 217: Unrelated noise? Should we omit it from the PR or keep it in?
Personally, I like the following strategy:
- Have dedicated commits for config noise that has no functional consequence and that we are sure is reproducible.
  - Technically they would be "out of scope", but it is simply not practical to create dedicated issues.
  - This can even be split further into "reorder config keys" vs other noise.
  - One way to produce separate commits is to run config export _before_ doing the functional change. In this case, export the config and commit, before we enable ckeditor5.
- Do not include config noise that is not reproducible (e.g. depends on local env or external randomness)
- Do not include config noise that would cause functional changes or have some kind of risk. Instead, open separate issues.

tests/src/ExistingSiteJavascript/WysiwygEmbedTest.php
```diff
@@ -90,18 +89,18 @@ public function testMediaEmbed(): void {
...
     $this->getSession()->switchToIFrame();
-    $assert_session->assertWaitOnAjaxRequest();
+    $assert_session->waitForElement('css', 'div.oe-oembed-entities-select-dialog');
+
     // Only one bundle is configured for embed.
     $assert_session->fieldNotExists('Display as');
-    $modal_button_pane = $assert_session->elementExists('css', 'div.ui-dialog-buttonpane');
+    // There are two embed dialogs, the second was the embed button.
+    $assert_session->waitForElementVisible('css', 'form.oe-oembed-entity-dialog-step--embed');
+    $modal_button_pane = $assert_session->elementExists('css', 'div.ui-dialog-buttonset');
```
donquixote, line 98: Are some of these wait statements redundant?
(I don't think it is a blocker, just asking)

tests/src/ExistingSiteJavascript/WysiwygEmbedTest.php
```diff
@@ -90,18 +89,18 @@ public function testMediaEmbed(): void {
...
     $modal_button_pane->findButton('Embed')->press();
-    $assert_session->assertWaitOnAjaxRequest();
 
     // Check that the element has been inserted in the editor.
-    $this->getSession()->switchToIFrame('ckeditor');
     $assert_session->linkExists('Document title');
```
donquixote, line 102: Interesting that this works without the wait above.
(ok for me)

config/install/editor.editor.full_html.yml
donquixote, line 1: This file is a bit special.

Before this PR:
The `filter.format.full_html` was already disabled, but the `editor.editor.full_html` was still enabled.
On re-enabling the text format, we see it was using ckeditor 4.

In a previous version of this PR:
Running updb would disable ckeditor module, which would remove the editor.editor.full_html (and any other editor config that used ckeditor 4).
Then the same updb would restore the new version of editor config for rich_text and simple_rich_text but not for full_html.
As a consequence, editor.editor.full_html was now missing in the database config, but still present in the config/install of oe_showcase, and in oe_showcase.info.yml.

In the new version of the PR:
I enabled the text format, switched the editor to ckeditor5, then disabled the text format again.

donquixote (commented):
I think all or most of this PR is good.

Also, for a regular website that moves to ckeditor5, they need to check how the editor will affect their existing content. But for oe_showcase we don't need to worry about this. Our default content does not have advanced markup like tables.

The next reviewer needs to check my latest commits.

drishu (commented):
Changed look good

donquixote (dismissed):

tibi2303 (approved):

PR #313, opened 2025-06-16
OEL-3893: Configure OpenEuropa Unified eTrans submodule

piotrsmykaj:


modules/oe_showcase_multilingual/oe_showcase_multilingual.module
AaronGilMartinez, line 1: There is nothing in the file so no need to have it.

AaronGilMartinez: Tests are failing, since behaviour of the switcher changed they will need to be updated too.

composer.lock
```diff
@@ -7880,32 +7880,31 @@
         },
         {
             "name": "openeuropa/oe_dashboard_agent",
-            "version": "1.1.0",
```
AaronGilMartinez: I see that oe_dashboard_agent is updated but not sure if there is a relation with the changes.

AaronGilMartinez (changes requested):

composer.lock
piotrsmykaj: Rebased, and now it doesn't happen

modules/oe_showcase_multilingual/oe_showcase_multilingual.module
piotrsmykaj: fixed both remarks, thanks

AaronGilMartinez (approved):

PR #315, opened 2025-06-18
OEL-3912: Replace Seven with Claro.

drishu:
### Description

Replace seven with claro as there is not hard dependency or requirements for it and claro is the recommended theme for replacing it.

### Change log

- Changed: Admin theme & config
- Removed: Seven theme config


oe_showcase.info.yml
```diff
@@ -224,14 +224,14 @@ config_devel:
...
-    - block.block.seven_content
-    - block.block.seven_help
-    - block.block.seven_local_actions
-    - block.block.seven_messages
-    - block.block.seven_page_title
-    - block.block.seven_primary_local_tasks
-    - block.block.seven_secondary_local_tasks
+    - block.block.claro_breadcrumbs
+    - block.block.claro_content
+    - block.block.claro_help
```
AaronGilMartinez: I don't see block.block.claro_help

AaronGilMartinez (changes requested):

oe_showcase.info.yml
drishu: thx

AaronGilMartinez (approved):

PR #319, opened 2025-07-01
EWPP-4664: Update to Drupal 11.1

peterserfozo:
### Description

Update to Drupal 11.1


composer.json
```diff
@@ -14,7 +14,7 @@
         "drupal/block_field": "^1.0@RC",
         "drupal/color_scheme_field": "^1.0@beta",
         "drupal/composite_reference": "^2.4",
-        "drupal/core-recommended": "^10",
+        "drupal/core-recommended": "^11.1 <11.2",
```
intelektron, line 17: We don't allow D10 anymore?
The profile info file says `core_version_requirement: ^10 || ^11`

intelektron (approved):

composer.json
```diff
@@ -69,9 +68,9 @@
         "drupal/cas_mock_server": "^2.0.2",
         "drupal/config_devel": "^1.10",
         "drupal/core-composer-scaffold": "^10",
-        "drupal/core-dev": "^10",
+        "drupal/core-dev": "^11.1 <11.2",
```
AaronGilMartinez, line 71: I think scaffold is missing the update.

AaronGilMartinez (changes requested):

composer.json
peterserfozo: Fixed, thanks!

AaronGilMartinez (approved):

PR #321, opened 2025-07-01
OEL-3939: Remove ec_core_column footer section.

enriquelacoma:


AaronGilMartinez (changes requested):
Since we are altering the regular behaviour of the module I would add checks in the tests to see that this links are not displayed.

tests/src/ExistingSite/DefaultFooter.php
```diff
@@ -0,0 +1,30 @@
...
+class DefaultFooter extends ShowcaseExistingSiteTestBase {
+
+  /**
+   * Test ec-core column is removed.
+   */
+  public function testNoEcCoreColumn(): void {
+
+    $this->drupalGet('');
+    $assert = $this->assertSession();
+    $this->assertSession()->elementTextNotContains('css', 'footer', 'EC Core column');
```
AaronGilMartinez: You can use the $assert variable here.

tests/src/ExistingSite/DefaultFooter.php
AaronGilMartinez, line 1: Right now we have two tests for the footer.
https://github.com/openeuropa/oe_showcase/blob/1.x/tests/src/ExistingSite/ManageSiteSpecificFooterTest.php 

We could merge into one with two methods: 
- Check the links management.
- Check that EC Core links are not displayed.

modules/oe_showcase_navigation/oe_showcase_navigation.post_update.php
```diff
@@ -21,3 +21,21 @@ function oe_showcase_navigation_post_update_00001(&$sandbox): void {
...
+    'oe_corporate_blocks.footer_link.general.funding_and_tenders',
+    'oe_corporate_blocks.footer_link.general.jobs',
+    'oe_corporate_blocks.footer_link.general.press_corner',
+    'oe_corporate_blocks.footer_link.general.priorities',
+    'oe_corporate_blocks.footer_link.general.topics',
+  ];
+  foreach ($config as $item) {
+    \Drupal::configFactory()->getEditable($item)?->delete();
+  }
+  \Drupal::configFactory()->getEditable('oe_corporate_blocks.footer_link.section.ec_core_column')?->delete();
```
AaronGilMartinez: Why not adding this to the array? Same for the install

AaronGilMartinez (changes requested):

tests/src/ExistingSite/DefaultFooter.php
enriquelacoma: Done

tests/src/ExistingSite/DefaultFooter.php
enriquelacoma: Done

modules/oe_showcase_navigation/oe_showcase_navigation.post_update.php
enriquelacoma: Done

AaronGilMartinez (approved):

PR #334, opened 2025-08-05
OEL-3724: Added test for aria-label content multiselect.

tibi2303:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/recipes/.gitignore
+++ b/recipes/.gitignore
@@ -1,0 +1,1 @@
+/README.txt
--- a/tests/src/ExistingSiteJavascript/MultiSelectAccessibleTest.php
+++ b/tests/src/ExistingSiteJavascript/MultiSelectAccessibleTest.php
@@ -30,6 +30,10 @@ class MultiSelectAccessibleTest extends ShowcaseExistingSiteJavascriptTestBase {
 
     $this->assertEquals('false', $multiselect->getAttribute('aria-expanded'));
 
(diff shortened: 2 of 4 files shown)


recipes/.gitignore
AaronGilMartinez, line 1: Doesn't seem related to the changes, but if we don't provide recipes we shouldn't add the folder. Maybe add it to .gitignore.

tests/src/ExistingSiteJavascript/MultiSelectAccessibleTest.php
```diff
@@ -30,6 +30,11 @@ public function testSlimSelectAriaAttributes(): void {
 
     $this->assertEquals('false', $multiselect->getAttribute('aria-expanded'));
 
+    $multiselectList = $page->find('css', '.ss-list');
+    $this->assertNotEmpty($multiselectList->getAttribute('aria-label'));
+    $this->assertEquals('Select an option', $multiselectList->getAttribute('aria-label'));
+    $this->assertEquals('listbox', $multiselectList->getAttribute('role'));
```
AaronGilMartinez, line 36: If slim select does this check in the tests we don't need to check here.

AaronGilMartinez (changes requested):

PR #340, opened 2025-09-23
OEL-4016: Mega menu

donquixote:


modules/oe_showcase_example_route/oe_showcase_example_route.info.yml
piotrsmykaj, line 1: I am wondering why duplicating this module with `oe_bootstrap_theme_route_test` as they do the same thing. If it's just a matter of url /showcase/ vs /oebt/ for me it should not be an issue because it's purposed just for example content usage.

piotrsmykaj: @donquixote explained that the one from OEBT should not be available on the websites as it's only for tests purposes.

piotrsmykaj (dismissed):

piotrsmykaj (dismissed):

tibi2303 (approved):

PR #341, opened 2025-09-24
OEL-4182: Use latest stable release tag in drone "update" step.

donquixote:


.drone.yml
```diff
@@ -57,7 +57,14 @@ pipeline:
     commands:
       - git config --global --add safe.directory /test/oe_showcase
       - git fetch --tags
-      - git checkout $(git describe --tags $(git rev-list --tags --max-count=1))
+      - |
+        set -ex
+        # Find the latest stable release tag in the history of the current commit.
+        # Exclude *-alpha* and *-beta* tags, and also tmp releases like 0.****.********.
+        # Abort if no such release found.
+        LATEST_TAG=$(git describe --abbrev=0 --tags --match=[0-9]*.[0-9]*.[0-9]* --exclude=*.*.???*)
```
AaronGilMartinez: If I understood the rules, this will allow semantic versions, excluding some special cases.

Questions/points about this:
1. If we want to pick only stables, shouldn't it be [1-9]*.[0-9]*.[0-9]*? Any release starting with 0.x is temp.
2. Not sure if the exclusion is needed in this case. I think we don't use alpha/betas for this package, it is mainly used in OESA and we have only temp tags.

AaronGilMartinez: Once we agree, I think we should extend this to gitlab CI (if not already like this).

donquixote: > If we want to pick only stables, shouldn't it be [1-9].[0-9].[0-9]*? Any release starting with 0.x is temp.

Good point!
This will include all -alpha and -beta releases, but that's ok.

> Not sure if the exclusion is needed in this case.

It is not anymore.

> Once we agree, I think we should extend this to gitlab CI (if not already like this).

The gitlab is not configured yet. This will happen in a new ticket.

AaronGilMartinez (approved):

PR #346, opened 2025-09-30
OEL-4179: Use drupal/core:~11.2.0.

donquixote:


composer.json
```diff
@@ -16,7 +16,7 @@
         "drupal/block_field": "^1.0@RC",
         "drupal/color_scheme_field": "^1.0@beta",
         "drupal/composite_reference": "^2.4",
-        "drupal/core-recommended": "^11.1 <11.2",
+        "drupal/core-recommended": "~11.2.0",
```
tibi2303, line 18: can we have it 11.2, not 11.2.0

donquixote: This is oe_showcase, we can be more specific.
With `~11.2.0` we allow all of `11.2.*` but we don't allow any of `11.3.*` or later.
With `~11.2` we would also allow `11.3.*` or later.

runner.yml.dist
```diff
@@ -69,6 +69,13 @@ commands:
         $config['symfony_mailer.mailer_transport.sendmail']['configuration']['pass']='';
         $config['symfony_mailer.mailer_transport.sendmail']['configuration']['host']='mailpit';
         $config['symfony_mailer.mailer_transport.sendmail']['configuration']['port']='1025';
+
+        // Enable full error reporting.
+        error_reporting(E_ALL);
+        ini_set('display_errors', TRUE);
+        ini_set('display_startup_errors', TRUE);
+        $config['system.logging']['error_level'] = 'verbose';
+
```
donquixote: This change is not absolutely required, we could remove it from the PR.
TBD.

docker-compose.yml
```diff
@@ -9,7 +9,6 @@ services:
       # - nfsmount:/var/www/html # Mac Users with the nfsmount volume.
     environment:
       KEY_AI_GPT_AT_EC: ""
-      COMPOSERVER: "--2"
```
donquixote, line 12: We are removing this line because of https://git.fpfis.tech.ec.europa.eu/digit/devsecops/httpd-php/-/issues/3

composer.json
```diff
@@ -48,29 +48,29 @@
         "microsoft/powerbi": "^2.1",
         "openeuropa/composer-artifacts": "^1.0.0-alpha1",
         "openeuropa/oe_agenda": "^1.0.0-alpha4",
-        "openeuropa/oe_ai_provider_gpt_at_ec": "^1.0@alpha",
+        "openeuropa/oe_ai_provider_gpt_at_ec": "^1.0.0-alpha3",
```
donquixote, line 51: This explicit require is needed, otherwise composer will downgrade this package to alpha1 so that other packages can be upgraded.

runner.yml.dist
AaronGilMartinez: I think its better to move it to a separate command or drop it, does not make much sense to have it inside sendmail.

donquixote: wow indeed, I did not even realize this is a "sendmail" section.\
Going to drop it.

AaronGilMartinez (approved):

PR #347, opened 2025-10-14
OEL-3794: Install and configure AI modules

enriquelacoma:


composer.json
```diff
@@ -9,6 +9,9 @@
         "php": ">=8.1",
         "cweagans/composer-patches": "^1.7",
         "drupal/address": "^2.0",
+        "drupal/ai": "^1.1",
```
AaronGilMartinez: Is this needed? Probably this is main dependency for the other modules.

modules/oe_showcase_ai/oe_showcase_ai.info.yml
```diff
@@ -0,0 +1,34 @@
...
+  - drupal:ai_assistant_api
+  - drupal:ai_automators
+  - drupal:ai_chatbot
+  - drupal:ai_ckeditor
+  - drupal:ai_content_suggestions
+  - drupal:ai_external_moderation
+  - drupal:ai_image_alt_text
+  - drupal:ai_logging
+  - drupal:ai_translate
+  - openeuropa:oe_ai_provider_gpt_at_ec
```
AaronGilMartinez: {project}:{module}
All contribs will use their project machine name like `key:key`.
Only core modules use drupal.
Same for oe_ai_provider_gpt_at_ec.

modules/oe_showcase_ai/oe_showcase_ai.info.yml
```diff
@@ -0,0 +1,34 @@
+name: OpenEuropa Showcase AI
+type: module
+description: Provides the configuration for AI services.
+core_version_requirement: ^10 || ^11
```
AaronGilMartinez: I don't think we need to support Drupal 10 since the version is locked.

AaronGilMartinez (changes requested):
For the tests we could add some functional tests and mock the provider, but the thing is: Does it make sense? This would be testing module functionality with specific configuration.

Another issue here is that ai_ckeditor_extras and ai_image_alt_text don't have test at all.

Other option is to check configuration applied, but is too much and the package does not do this anywhere else.

In conclusion I think some basic functional tests wouldn't hurt even if its the module's job, the idea of existing site tests is to check site functionality.

We can mock the provider for this as the AI module or the library does.

modules/oe_showcase_ai/oe_showcase_ai.info.yml
enriquelacoma: Done

composer.json
enriquelacoma: Removed

modules/oe_showcase_ai/oe_showcase_ai.info.yml
enriquelacoma: Removed

enriquelacoma:
> For the tests we could add some functional tests and mock the provider, but the thing is: Does it make sense? This would be testing module functionality with specific configuration.
> 
> Another issue here is that ai_ckeditor_extras and ai_image_alt_text don't have test at all.
> 
> Other option is to check configuration applied, but is too much and the package does not do this anywhere else.
> 
> In conclusion I think some basic functional tests wouldn't hurt even if its the module's job, the idea of existing site tests is to check site functionality.
> 
> We can mock the provider for this as the AI module or the library does.

I added test that check the plugins are enabled when creating a node, even if they wont work because there is no connection to the ai, the provider module doesn't not have test, so I think it is overkill to mock for test it in the site.

AaronGilMartinez (approved):

